PR #9, opened 2023-02-07
Apollo list view POC (can be closed, but left it open for future reference)

acidjunk:


apps/example-orchestrator-ui/pages/subscriptions.tsx
```diff
@@ -0,0 +1,90 @@
...
+    edges {
+      node {
+        note
+        name
+        startDate
+        endDate
+        tag
+        productId
+        portSubscriptionInstanceId
+        vlanRange
```
pboers1988, line 21: SURF specific? I guess you are implementing our internal API however that version of the api will never be published. I guess I might be missing some discussions over the past week but how will this become generic?

acidjunk: If I am not mistaken: there are stories on the sprint: to add GraphQL to orchestrator-core? So when UI 2.0 is ready; we should have generic GraphQL access, right?

For the time being: I want to know what pitfalls we might encounter: and try to tackle Type issues (that arise when other users implement different backends.

PR #13, opened 2023-02-10
Husky: pre-commit Linting and Prettier check

ricardovdheijden:


tjeerddie (changes requested):
set husky as executable.
```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
```

tjeerddie (approved):

PR #45, opened 2023-03-22
Feature/navigation environment label

ricardovdheijden:
- Customizable environment text (production is light-blue, other texts result in yellow label)
- Moving each badge into its own component
- Introducing helper function to make colors lighter to be a secondary color (mix with white 80%)
- Making HeaderBadge accept also custom colors (string) while maintaining the auto complete for the theme colors
- Fallback text when enginestatus is unavailable

image
image
image


libs/orchestrator-ui-components/src/lib/components/OrchestratorPageTemplate/OrchestratorPageHeader.tsx
```diff
@@ -74,22 +81,17 @@ export const OrchestratorPageHeader: FC<OrchestratorPageHeaderProps> = ({
...
-                        <HeaderBadge
-                            color="emptyShade"
-                            iconType={() => (
-                                <StatusDotIcon color={theme.colors.success} />
-                            )}
-                        >
-                            Engine is {engineStatus?.global_status}
-                        </HeaderBadge>
-                        <FailedTasksBadge {...taskCountsSummary} />
+                        <EngineStatusBadge engineStatus={engineStatus} />
```
tjeerddie, line 91: Since you made it specific for engineStatus, use `useEngineStatusQuery` inside the component instead of passing it from here.

pboers1988:
LGTM

PR #60, opened 2023-05-01
Table component and partial implementation of subscriptions page

ricardovdheijden:
Table features:
- Typesafe link between data and column definition
- The props in CellRender method knows the data type of the data in the cell
- Non-data colums, like the "+" in the subscription table

The subscription page contains an implementation of the Table component showing subscriptions. Note: the pagination is not accurate yet


apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
```diff
@@ -0,0 +1,266 @@
...
+        tag: {
+            initialWidth: 100,
+            displayAsText: 'Tag',
+        },
+        startDate: {
+            displayAsText: 'Start Date',
+            initialWidth: 150,
+            renderCell: (cellValue) =>
+                cellValue
+                    ? new Date(parseInt(cellValue) * 1000).toLocaleString(
```
acidjunk, line 105: is this timezone aware, e.g. CEST for us?

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
```diff
@@ -0,0 +1,266 @@
...
+                endDate,
+                status,
+                subscriptionId,
+                notes: note,
+            };
+        },
+    );
+}
+
+function mapPythiaSortOrderToSortDirection(
```
acidjunk, line 260: I would suggest not adding "Pythia" that's an implementation detail for SURF (for now, until the orchestrator graphql stuff is ready).

libs/orchestrator-ui-components/src/lib/hooks/useQueryWithGraphql.ts
```diff
@@ -0,0 +1,27 @@
+import { useContext } from 'react';
+import { OrchestratorConfigContext } from '../contexts/OrchestratorConfigContext';
+import { GraphQLClient } from 'graphql-request';
+import { useQuery } from 'react-query';
+import { TypedDocumentNode } from '@graphql-typed-document-node/core';
+import { Variables } from 'graphql-request/build/cjs/types';
+
+export const useQueryWithGraphql = <U, V extends Variables>(
```
acidjunk, line 8: Nice!

acidjunk (dismissed):
Some small comment. Feel free to ignore some.

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
ricardovdheijden: Good one, done

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
ricardovdheijden: Added a todo in the code for now

Mark90: It's probably not localized :) Would be good to ticketify this todo so we don't find out about it next winter

apps/example-orchestrator-ui/components/Subscriptions/subscriptionsQuery.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+export const GET_SUBSCRIPTIONS_PAGINATED_DEFAULT_VARIABLES: SubscriptionGridQueryVariables =
+    {
+        first: 20,
+        after: 0,
+        sortBy: DEFAULT_SUBSCRIPTIONS_SORT_ORDER,
+    };
+
+export const getPythiaSortOrderFromString = (
```
Mark90, line 68: Pythia will soon be no more, also it's not a publicly available app; rename to "Graphql" (a todo is fine of course)

libs/orchestrator-ui-components/src/lib/components/Table/Table.tsx
```diff
@@ -0,0 +1,143 @@
...
+    getInitialColumnOrder,
+    TableColumns,
+} from './columns';
+import {
+    EuiDataGridControlColumn,
+    EuiDataGridPaginationProps,
+} from '@elastic/eui/src/components/datagrid/data_grid_types';
+
+// Total height of grid button bar, table header and pagination bar
+const EUI_DATA_GRID_HEIGHT_OFFSET = 103;
```
Mark90, line 23: This is not a height that's relevant to the multiplier? (i remember you showed that in one of the first meetings about the project, but forgot the exact name)

Mark90 (approved):

apps/example-orchestrator-ui/components/Subscriptions/subscriptionsQuery.ts
ricardovdheijden: True, when Pythia is no more, the return type of this function will change too. Once that happens I will update the name of this function too.

PR #62, opened 2023-05-11
Migrate nx to version 16

tjeerddie:


migrations.json
```diff
@@ -0,0 +1,204 @@
+{
```
ricardovdheijden, line 1: Is this something we need to commit?

ricardovdheijden (approved):

PR #100, opened 2023-06-11
POC of breadcrumb navigation

Georgi2704:


libs/orchestrator-ui-components/src/lib/components/breadcrumb-navigation/breadcrumb-navigation.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useCallback, useEffect, useState } from 'react';
+import { EuiBreadcrumb, EuiBreadcrumbs } from '@elastic/eui';
+import { NextRouter } from 'next/router';
```
acidjunk, line 3: Hmm, we can't use Next Router from inside the lib.
Some users will want to use the lib without Next.

I will probably refactor this, after merge. So just a reminder.

Issue #54, opened 2023-04-06
Implement settings page


hanstrompert:
Implement settings page with:
- add bread crumb, 
- add `Settings` header 
- Flush settings
- Modify engine settings
- Worker type
  - If no celery workers are being by the orchestrator deployment, don't display this last `Worker Status` information "component".

Base content and action of this page on the existing orchestrator. The Worker type design is missing in Invision, but use the following illustration:

![image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/c702a8ba-9d6e-4fd1-b84e-5b13f62e456d)

Invision prototype:
- https://projects.invisionapp.com/share/MW134V4MHPB7?hideLogo=true#/screens 

Sketch file (inspect-element & downloadable)
- https://www.sketch.com/s/68008066-6d55-4e28-8ba9-cebb8489fc95

acidjunk:
@hanstrompert : External?

acidjunk:
Leftovers:

- cache flush isn’t complete 
- the engine status at the top of the page should also toggle (observer can’t be used because the queryProvider lives in the app and not in the lib: not sure how to handle this correctly yet)
- the last comment about Celery workers isn't implemented yet

PR #104, opened 2023-06-17
Settings page

Georgi2704:
#54

### DIFF
--- a/libs/orchestrator-ui-components/src/lib/components/SettingsPage/EngineStatusButton.tsx
+++ b/libs/orchestrator-ui-components/src/lib/components/SettingsPage/EngineStatusButton.tsx
@@ -1,0 +1,46 @@
+import { EuiButton } from '@elastic/eui';
+import React from 'react';
+
+interface EngineStatusButtonProps {
+    isLoading: boolean;
+    isRunning: boolean;
+    changeEngineStatus: () => void;
+}
+
+export const EngineStatusButton = ({
+    isLoading,
+    isRunning,
+    changeEngineStatus,
+}: EngineStatusButtonProps) => {
+    if (isLoading) {
+        return (
+            <EuiButton isLoading fill>
+                Loading...
+            </EuiButton>
+        );
+    } else {
+        if (isRunning) {
+            return (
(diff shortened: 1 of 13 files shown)


libs/orchestrator-ui-components/src/lib/components/SettingsPage/EngineStatusButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+    isRunning,
+    changeEngineStatus,
+}: EngineStatusButtonProps) => {
+    if (isLoading) {
+        return (
+            <EuiButton isLoading fill>
+                Loading...
+            </EuiButton>
+        );
+    } else {
```
ricardovdheijden, line 21: Multiple levels of if-else might be confusing to read. Suggesting to remove the whole else block and rewrite it like this:

`return isRunning ? <EuiButtonFromLine24> : <EuiButtonFromLine35>`

libs/orchestrator-ui-components/src/lib/components/SettingsPage/Status.tsx
```diff
@@ -0,0 +1,70 @@
...
+    return (
+        <EuiPanel
+            hasShadow={false}
+            color="subdued"
+            paddingSize="l"
+            style={{ width: '50%' }}
+        >
+            <EuiFlexGroup>
+                <EuiFlexItem grow={false}>
+                    <EuiText size="s" style={{ fontWeight: 600 }}>
```
ricardovdheijden, line 29: Fontweights are part of the theme. You can use the value from the theme here instead of hardcoding.
`libs/orchestrator-ui-components/src/lib/theme/defaultOrchestratorTheme.ts`

libs/orchestrator-ui-components/src/lib/components/SettingsPage/EngineStatusButton.tsx
acidjunk: I will fix this

acidjunk:
Added some leftovers to #54

acidjunk (approved):

PR #109, opened 2023-06-21
Feature/table filter

ricardovdheijden:
Adds filter field to the Subscriptions component. https://github.com/workfloworchestrator/orchestrator-ui/issues/76

### DIFF
--- a/apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
+++ b/apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
@@ -21,7 +21,7 @@ import {
 import React, { FC } from 'react';
 import { useRouter } from 'next/router';
 import Link from 'next/link';
-import { EuiFlexItem } from '@elastic/eui';
+import { EuiFlexItem, EuiSearchBar } from '@elastic/eui';
 import {
     DESCRIPTION,
     END_DATE,
@@ -71,6 +71,8 @@ export type SubscriptionsProps = {
     setPageIndex: (updatedPageIndex: number) => void;
     sortOrder: SubscriptionsSort;
     setSortOrder: (updatedSortOrder: SubscriptionsSort) => void;
+    filterQuery: string;
+    setFilterQuery: (updatedFilterQuery: string) => void;
 };
 
 export const Subscriptions: FC<SubscriptionsProps> = ({
@@ -80,6 +82,8 @@ export const Subscriptions: FC<SubscriptionsProps> = ({
     setPageSize,
     setPageIndex,
     setSortOrder,
+    filterQuery,
+    setFilterQuery,
 }) => {
     const router = useRouter();
     const { theme } = useOrchestratorTheme();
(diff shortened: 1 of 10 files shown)


libs/orchestrator-ui-components/src/lib/components/SearchBar/searchQueryMapper.ts
```diff
@@ -0,0 +1,48 @@
...
+    queryContainer: QueryContainer,
+): [string, string] | undefined {
+    if (queryContainer.match !== undefined) {
+        const firstKey: string = Object.keys(queryContainer.match)[0];
+
+        const firstValue = (
+            queryContainer.match[
+                firstKey as keyof typeof queryContainer.match
+            ] as MatchObject
+        ).query;
```
mrijk, line 18: <rant>

It feels a bit that something has gone terribly wrong in the industry if this is needed to get a simple match ;) Not even sure if this is inherently type safe or if it's just stating that it is typesafe by all this declaring/casting.

If I understand this correctly 'firstKey' is a string, and then becomes kinda typesafe by declaring that it is a key in this match. Which it was in the first place, since you got the key form the queryContainer match. But then again, I don't know enough of typescript to understand if this can be avoided.

I would have expected a oneliner, but got 10 lines instead.

</rant>

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
```diff
@@ -200,10 +216,9 @@ export const Subscriptions: FC<SubscriptionsProps> = ({
         });
 
     const handleCriteriaChange = (criteria: Criteria<Subscription>) => {
-        if (criteria.page) {
-            const {
-                page: { index, size },
-            } = criteria;
+        const { page } = criteria;
```
mrijk, line 219: No desctructuring in the parameter?

`const handleCriteriaChange = ({ page }: Criteria<Subscription>) =>`

libs/orchestrator-ui-components/src/lib/components/SearchBar/searchQueryMapper.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    // returning undefined for unsupported query-matchers
+    return undefined;
+}
+
+export function isValidQueryPart(
+    filter: string[] | undefined,
+): filter is [string, string] {
+    return filter !== undefined && filter?.length === 2;
```
mrijk, line 47: Is this check for undefined needed if your use the ?. operator?

libs/orchestrator-ui-components/src/lib/components/SearchBar/searchQueryMapper.ts
ricardovdheijden: This piece of code I am least proud of indeed. The problem is that when you use Object.keys you get an array of strings instead of an array of "keyof ..." therefore I used the typecast here. It gets even worse when I also need to cast the result to be able to access the query prop. Another solution I thought about is just adding a ts-ignore statement.

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
ricardovdheijden: Yes, good one

libs/orchestrator-ui-components/src/lib/components/SearchBar/searchQueryMapper.ts
ricardovdheijden: Indeed too much code here, the last part is indeed sufficient

mrijk (approved):

PR #117, opened 2023-06-26
Feature/subscription overview tabs

ricardovdheijden:
Adding tabs to the subscriptions overview. Every tab sets a fixed filter on the status of the subscriptions.

https://github.com/workfloworchestrator/orchestrator-ui/issues/77


libs/orchestrator-ui-components/src/lib/components/SubscriptionsTabs/getSubscriptionsTabTypeFromString.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    it('returns TRANSIENT', () => {
+        const result = getSubscriptionsTabTypeFromString(
+            SubscriptionsTabType.TRANSIENT,
+        );
+        expect(result).toEqual(SubscriptionsTabType.TRANSIENT);
+    });
+
+    it('returns ALL', () => {
```
tjeerddie, line 26: also test normal string instead of only `SubscriptionsTabType`, `"all"` should also work right?

libs/orchestrator-ui-components/src/lib/components/SubscriptionsTabs/SubscriptionsTabs.tsx
```diff
@@ -0,0 +1,69 @@
...
+    selectedSubscriptionsTab,
+    onChangeSubscriptionsTab,
+}) => {
+    return (
+        <EuiTabs>
+            {tabs.map(({ id, name }) => (
+                <EuiTab
+                    key={id}
+                    isSelected={id === selectedSubscriptionsTab}
+                    onClick={() => {
```
tjeerddie, line 58: I'd like to keep as much logic out of the template, move it to a const

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
```diff
@@ -180,7 +182,11 @@ export const Subscriptions: FC<SubscriptionsProps> = ({
     );
     const queryContainsInvalidParts =
         filterQueryTupleArray?.includes(undefined);
-    const filterBy = filterQueryTupleArray?.filter(isValidQueryPart);
+    const filterBy =
+        filterQueryTupleArray
+            ?.concat([alwaysOnFilter])
+            .filter(isValidQueryPart) ||
+        (alwaysOnFilter ? [alwaysOnFilter] : undefined);
```
tjeerddie, line 189: is this check needed? empty array should basically be the same as undefined.

apps/example-orchestrator-ui/pages/subscriptions/index.tsx
```diff
@@ -74,6 +98,11 @@ export default function SubscriptionsPage() {
                 }}
                 filterQuery={filterQuery}
                 setFilterQuery={setFilterQuery}
+                alwaysOnFilter={
+                    defaultSubscriptionsTabs.find(
```
tjeerddie, line 102: move to a const outside the template.

libs/orchestrator-ui-components/src/lib/components/SubscriptionsTabs/getSubscriptionsTabTypeFromString.spec.ts
ricardovdheijden: You are right, the input must always be a string

apps/example-orchestrator-ui/pages/subscriptions/index.tsx
```diff
@@ -41,22 +51,36 @@ export default function SubscriptionsPage() {
...
         <NoSSR>
             <EuiSpacer />
+
             <EuiPageHeader pageTitle="Subscriptions" />
+            <EuiSpacer size="m" />
+
+            <SubscriptionsTabs
+                tabs={defaultSubscriptionsTabs}
+                selectedSubscriptionsTab={selectedSubscriptionsTab}
+                onChangeSubscriptionsTab={(updatedSubscriptionsTab) => {
```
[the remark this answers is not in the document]
ricardovdheijden: done

apps/example-orchestrator-ui/pages/subscriptions/index.tsx
ricardovdheijden: done

libs/orchestrator-ui-components/src/lib/components/SubscriptionsTabs/SubscriptionsTabs.tsx
ricardovdheijden: done

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
tjeerddie: ohh, I see, I do believe it can be improved by removing the special undefined logic. change alwaysOnFilter to have default `alwaysOnFilter=[]` and not be undefined, the backend is fine with an empty array.

tjeerddie: It would be better to change the `alwaysOnFilter` to accept multiple filters

DutchBen (approved):

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
ricardovdheijden: done, supporting multiple alwaysOnFilters now too

tjeerddie (approved):

PR #121, opened 2023-06-29
gitlab-19: Use yalc to support local development on gpl lib and surf app at the same time

DutchBen:
## Summary
Setup to use Yalc as a tool to be able to work on the GPL component library and the Surf app implemation locally at the same time

## DOD Tasks
X Add/Update documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,10 @@ Thumbs.db
 
 # Next.js
 .next
+
+# VsCode
+.vscode
+
+# Yalc
+/.yalc
+yalc.lock
--- a/.gitignore
+++ b/.gitignore
@@ -19,10 +19,6 @@ node_modules
 
 # IDE - VSCode
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -41,3 +41,10 @@ Thumbs.db
 
 # Next.js
 .next
+
+# VsCode
+.vscode
```
ricardovdheijden, line 46: If I am correct VS Code is covered above at line 21

ricardovdheijden (dismissed):

.gitignore
DutchBen: You are correct! I missed that. The reason I readded it is it's trying to commit .vscode/settings.json, that file get's excluded from the ignore at line 21. What do you think, should we include the file or not? It currently contains 

```{
  "cSpell.words": [
    "Yalc"
  ]
}```

DutchBen: The solution for this was to remove all exceptions to the .vscode folder from the .gitignore

ricardovdheijden (approved):

PR #123, opened 2023-07-01
Improve breadcrumbs with next router

acidjunk:
Next Router + Docs + ADR's

Do we need to add docs for SURF specific?

Client app currently needs:
- react 18, strict mode off
- transpile options

```
const nextConfig = {
    reactStrictMode: false,
    transpilePackages: ['@orchestrator-ui/orchestrator-ui-components'],
};
```


libs/orchestrator-ui-components/src/lib/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,80 @@
...
+                router.push('/');
+            },
+        },
+    ];
+
+    // Handle remaining breadcrumbs dynamic
+    const parts = router.asPath.split('/');
+    parts.forEach((p, index) => {
+        if (index > 0) {
+            let link = parts.slice(0, index + 1).join('/');
```
mrijk, line 24: Not sure about live in Java/TypeScript land nowadays, but wouldn't it be clear to use a const and just assign to another const variable with a new meaningful name?

mrijk: Also you could maybe split up this quite big function in several smaller ones

libs/orchestrator-ui-components/src/lib/utils/string.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { upperCaseFirstChar } from './strings';
+
+describe('upperCaseFirstChar()', () => {
+    it("Doesn't crash on an empty string", () => {
```
ricardovdheijden, line 4: I would suggest to describe what the expected result is here.

libs/orchestrator-ui-components/src/lib/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,80 @@
...
+                href: link,
+                onClick: (e) => {
+                    e.preventDefault();
+                    if (
+                        text === 'Subscriptions' &&
+                        containsUuid4(router.asPath)
+                    ) {
+                        // Todo: make URLS more consistent or design a better way to handle breadcrumbs
+                        // When possible try to use the browser back; so the user has active tab + correct page
+                        router.back();
```
ricardovdheijden, line 45: This one back-fires (no pun intended) when you open a specific subscription detail in a new tab: http://localhost:4200/subscriptions/9f42c702-bed7-4c27-b8e8-05d446053065.

When you click on subscriptions in the breadcrumb you will end up in the browsers empty new-tab page.

libs/orchestrator-ui-components/src/lib/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,80 @@
...
+                text = upperCaseFirstChar(text.split('?').slice(0, -1).join());
+            }
+
+            breadcrumbs.push({
+                text: text,
+                href: link,
+                onClick: (e) => {
+                    e.preventDefault();
+                    if (
+                        text === 'Subscriptions' &&
```
DutchBen, line 40: Why is there a specific check for a string here. Can this be made more generic?

libs/orchestrator-ui-components/src/lib/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+            // Handle tabs if any
+            const tabs = parts.slice(0, index + 1).join('/');
+            if (tabs.includes('?activeTab=')) {
+                const tabName = upperCaseFirstChar(
+                    tabs.split('?activeTab=')[1].split('&')[0],
+                );
+                breadcrumbs.push({
+                    text: tabName,
+                    href: `/subscriptions?activeTab=${tabName}`,
```
DutchBen, line 60: Can this be more generic so it works for other usecases

mkdocs/mkdocs.yml
```diff
@@ -0,0 +1,43 @@
+site_name: Portal backend docs
```
DutchBen, line 1: What portal?

libs/orchestrator-ui-components/src/lib/components/Breadcrumbs/Breadcrumbs.tsx
acidjunk: Yes, refactored it. And wrote tests for the smaller functions.

libs/orchestrator-ui-components/src/lib/components/Breadcrumbs/Breadcrumbs.tsx
acidjunk: Removed this and floowed @wouter1975 specs 100% : no smart routing in the BreadCrumbs: only links to top level pages; So filter, tab and paginate will reset to defaults

libs/orchestrator-ui-components/src/lib/components/Breadcrumbs/Breadcrumbs.tsx
acidjunk: I was trying to make the Breadcrumbs smart enough to handle some special cases and testing the approach; reverted it.

libs/orchestrator-ui-components/src/lib/components/Breadcrumbs/Breadcrumbs.tsx
acidjunk: Not relevant anymore

mkdocs/mkdocs.yml
acidjunk: Fixed that: I now uses mkdocs for docs; because we already have som experience with it.
Not sure if thats a good choice...

Curious if you know some better options

tjeerddie (approved):

PR #132, opened 2023-07-05
88 fixed inputs page

acidjunk:


libs/orchestrator-ui-components/src/lib/hooks/useDataDisplayParams.ts
```diff
@@ -0,0 +1,64 @@
...
+export const useDataDisplayParams = <Type>(
+    initialParams: Partial<DataDisplayParams<Type>>,
+): DataDisplayReturnValues<Type> => {
+    const defaultParams = {
+        pageSize: DEFAULT_PAGE_SIZE,
+        pageIndex: 0,
+        ...initialParams,
+    };
+
+    const [dataDisplayParams, setDataDisplayParams] = useQueryParams({
```
ricardovdheijden, line 35: Just checking, is it not possible to specify the type "DataDisplayParams<Type>" directly after the useQueryParams?
`useQueryParams<DataDisplayParams<Type>>` to avoid the type cast in the return statement.

libs/orchestrator-ui-components/src/lib/components/Table/columns.ts
```diff
@@ -15,28 +15,23 @@ export type TableColumnsWithExtraNonDataFields<T> = TableColumns<T> & {
...
 
-export enum SortDirection {
-    Asc = 'ASC',
-    Desc = 'DESC',
-}
-
 export type DataSorting<T> = {
-    columnId: keyof T;
-    sortDirection: SortDirection;
+    columnId?: keyof T;
```
ricardovdheijden, line 19: For dataSorting I would say that both columnId and sortDirection should be always defined. In case of not sorting, make the whole prop DataSorting optional.

libs/orchestrator-ui-components/src/lib/components/Table/tableUtils.ts
```diff
@@ -1,18 +1,64 @@
...
+                dataDisplayParams.sortBy &&
+                dataDisplayParams.sortBy.order &&
+                dataDisplayParams.sortBy.field
+            ) {
+                return determineNewSortOrder<Type>(
+                    dataDisplayParams.sortBy?.field,
+                    dataDisplayParams.sortBy?.order,
+                    newSortField,
+                );
+            } else {
```
ricardovdheijden, line 42: "else" could be omitted here

apps/example-orchestrator-ui/components/Metadata/Products/Products.tsx
```diff
@@ -0,0 +1,170 @@
...
+            />
+        </>
+    );
+};
+
+function mapApiResponseToSubscriptionTableData(
+    graphqlResponse: ProductsResult,
+): Product[] {
+    return graphqlResponse.products.page.map((product): Product => {
+        const {
```
ricardovdheijden, line 150: I think this can be shortened to something like return { ... product }

libs/orchestrator-ui-components/src/lib/hooks/useDataDisplayParams.ts
DutchBen: I have tried to pass a type hint to useQueryParams so it would return the right thing right away but couldn't get that to work. Baring that I think the most elegant way is to typecast at the end since you see what going on a bit easier

libs/orchestrator-ui-components/src/lib/components/Table/columns.ts
DutchBen: Agreed. Is there a case when not sorting though? I have a default sorting on products of name/ASC now

libs/orchestrator-ui-components/src/lib/components/Table/tableUtils.ts
DutchBen: Removed

libs/orchestrator-ui-components/src/lib/components/Table/columns.ts
```diff
@@ -0,0 +1,38 @@
+import { EuiBasicTableColumn } from '@elastic/eui';
```
ricardovdheijden, line 1: I moved this file to libs/orchestrator-ui-components/src/lib/components/Table/utils/columns.ts

ricardovdheijden (approved):

apps/example-orchestrator-ui/components/Metadata/Products/Products.tsx
DutchBen: Yeah. I changed it. I'm not sure if I feel that's explicit enough though.

PR #133, opened 2023-07-05
flush cache and other additions to settings page

acidjunk:
Added ability to flush cache and header -> engineStatus now toggles when engineStatus changes.


libs/orchestrator-ui-components/src/lib/components/SettingsPage/FlushSettings.tsx
```diff
@@ -4,22 +4,68 @@ import {
...
+        }));
+    } else {
+        options = [{ label: 'Loading...' }];
+    }
+
+    const flushCache = () => {
+        if (selectedOptions.length < 1) {
+            return;
+        }
+        clearCache(
```
ricardovdheijden, line 48: This could be an async await for better readability

ricardovdheijden (approved):

PR #152, opened 2023-07-13
55 metadata products list

DutchBen:
- Searchfield is emitting an alert until full ES compatible search is enabled on the backend. The Subscriptions listing page is still working as it was but some dirty hacks where employed to make this both work (and not work)

- I have made a small beginning with renaming some badge components that I use with the WFO prefix.


apps/example-orchestrator-ui/components/Metadata/Products/Products.tsx
```diff
@@ -79,10 +83,27 @@ export const Products: FC<ProductsProps> = ({
...
             name: COLUMN_LABEL_STATUS,
+            width: '90',
+            render: (value) => (
+                <WFOStatusBadge status={value.toLocaleLowerCase()} />
+            ),
         },
         productBlocks: {
             field: PRODUCT_FIELD_PRODUCT_BLOCKS,
             name: COLUMN_LABEL_PRODUCT_BLOCKS,
+            render: (productBlocks) => {
```
tjeerddie, line 94: you don't need the function brackets when only having a return inside `(productBlocks) => (<></>)`. same with the function inside of this.

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
```diff
@@ -235,30 +229,24 @@ export const Subscriptions: FC<SubscriptionsProps> = ({
...
             pagination={pagination}
-            filterQuery={filterQuery}
-            isInvalidFilterQuery={queryContainsInvalidParts}
             isLoading={isFetching}
             localStorageKey={SUBSCRIPTIONS_TABLE_LOCAL_STORAGE_KEY}
-            onUpdateFilterQuery={setFilterQuery}
-            onCriteriaChange={handleCriteriaChange}
-            onUpdatePageSize={setPageSize}
-            onResetPageSize={() => setPageSize(defaultPageSize)}
+            onUpdatePageSize={(page) => {
```
tjeerddie, line 245: move this to a const function

libs/orchestrator-ui-components/src/lib/components/Table/TableWithFilter/TableWithFilter.tsx
```diff
@@ -135,7 +142,11 @@ export const TableWithFilter = <T,>({
                 onDataSort={onUpdateDataSort}
                 pagination={pagination}
                 isLoading={isLoading}
-                onCriteriaChange={onCriteriaChange}
+                onCriteriaChange={(criterion: Criteria<T>) => {
```
tjeerddie, line 145: move this to a const function

libs/orchestrator-ui-components/src/lib/components/Table/TableWithFilter/TableWithFilter.tsx
```diff
@@ -22,57 +22,57 @@ import {
...
-    isInvalidFilterQuery: boolean;
+    esQueryString?: string;
     isLoading: boolean;
     localStorageKey: string;
-    onUpdateFilterQuery: (updatedFilterQuery: string) => void;
-    onCriteriaChange: ({ page }: Criteria<T>) => void;
-    onUpdatePageSize: (updatedPageSize: number) => void;
-    onResetPageSize: () => void;
+    onUpdateEsQueryString: (esQueryString: string) => void;
+    onUpdatePageSize: (criterion: Criteria<T>['page']) => void;
```
ricardovdheijden, line 46: Suggesting to change the name "onUpdatePageSize" according to the data that is passed along. Page object contains the index and size.
Also maybe introduce an alias for Criteria<T>['page'] or even introduce a custom type without refering to Criteria

libs/orchestrator-ui-components/src/lib/components/StartPage/ListStartPage.tsx
```diff
@@ -15,7 +15,8 @@ interface ListStartPage {
 
 export default function ListStartPage({ list }: ListStartPage): ReactElement {
     return (
-        list && (
+        list &&
+        list.items && (
```
ricardovdheijden, line 19: Could shorten this to list?.items

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
```diff
@@ -235,30 +229,24 @@ export const Subscriptions: FC<SubscriptionsProps> = ({
...
-        if (page) {
-            const { index, size } = page;
-            setPageSize(size);
-            setPageIndex(index * size);
-        }
-    };
-
     return (
-        <TableWithFilter
+        <TableWithFilter<Subscription>
```
ricardovdheijden, line 233: Adding the type is not needed I would say. TS can determine the type based on de props

apps/example-orchestrator-ui/components/Metadata/Products/Products.tsx
```diff
@@ -122,24 +142,24 @@ export const Products: FC<ProductsProps> = ({
...
-                    setDataDisplayParam,
-                )}
-                pagination={pagination}
-                isLoading={isFetching}
-                onCriteriaChange={getPageSizeHandler<Product>(
-                    setDataDisplayParam,
-                )}
-            />
-        </>
+        <TableWithFilter<Product>
```
ricardovdheijden, line 145: Providing the type Product here is not needed, I would say.

apps/example-orchestrator-ui/components/Metadata/Products/Products.tsx
DutchBen: It's my personal preference to always include explicit returns. If you have to add something later - or temporarily like a console.log - it's easier. I'm as flexible though but I think we should probably make it a rule: 
https://eslint.org/docs/latest/rules/arrow-body-style, "as-needed" ?

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
DutchBen: What do you mean with a const function?

libs/orchestrator-ui-components/src/lib/components/Table/TableWithFilter/TableWithFilter.tsx
```diff
@@ -102,24 +102,31 @@ export const TableWithFilter = <T,>({
...
+            size: updatedTableConfig.selectedPageSize,
+        });
     };
 
     const handleResetToDefaults = () => {
         setHiddenColumns(defaultHiddenColumns);
         setShowSettingsModal(false);
         clearTableConfigFromLocalStorage(localStorageKey);
-        onResetPageSize();
+        onUpdatePageSize({
```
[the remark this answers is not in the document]
DutchBen: I have fixed it so it remembers the initial pageSize passed into it and reuses that

libs/orchestrator-ui-components/src/lib/components/Table/TableWithFilter/TableWithFilter.tsx
DutchBen: Updated the name of the function to onUpdatePage. Creating an alias is trickier with the generic being passed in, I think it's still nice to extend from the Criteria since it's supplied by EUI.

libs/orchestrator-ui-components/src/lib/components/StartPage/ListStartPage.tsx
DutchBen: fixed

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
DutchBen: Can't hurt either right?

apps/example-orchestrator-ui/components/Metadata/Products/Products.tsx
DutchBen: Can't hurt either right?

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
tjeerddie: what I mean is to move it out of the return and into a const above it.
```
const updatePageSize => {
  if (page) {
     setPageSize(page.size);
  }
}

return (<TableWithFilter onUpdatePageSize={updatePageSize} />
```

apps/example-orchestrator-ui/components/Metadata/Products/Products.tsx
tjeerddie: Good to make it a rule and you can always add the brackets when it becomes more then just a return.

DutchBen: made the rule and enforced it

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
DutchBen: fixed

libs/orchestrator-ui-components/src/lib/components/SearchBar/SearchField.tsx
```diff
@@ -1,31 +1,48 @@
...
+
+    return (
+        <EuiFormRow
+            fullWidth
+            isInvalid={!queryIsValid}
+            error={['The query contains invalid parts']}
+        >
+            <EuiSearchBar
+                query={queryString}
+                onChange={({ queryText }) => {
```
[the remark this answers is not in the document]
DutchBen: fixed

libs/orchestrator-ui-components/src/lib/components/Table/TableWithFilter/TableWithFilter.tsx
DutchBen: fixed

apps/example-orchestrator-ui/components/Subscriptions/Subscriptions.tsx
ricardovdheijden: True, but seeing this for example in plain text there might be a confusion if this is a <Subscription> component. Also I always understood that we should let typescript do its type-inference-thing as much as possible and only add types when TS is getting confused or when it is unclear.

ricardovdheijden (approved):

PR #160, opened 2023-07-19
Feature/112 update graphql url

ricardovdheijden:
Migrating from the Pythia graphql endpoint to the new core endpoint
https://github.com/workfloworchestrator/orchestrator-ui/issues/112

Addiditonal changes in this PR:
- Only using generated types from graphql
- Using the query parameter hook in Subscriptions page


apps/example-orchestrator-ui/components/Subscription/subscriptionQuery.ts
```diff
@@ -210,9 +145,12 @@ export function mapApiResponseToSubscriptionDetail(
         ),
         note: subscription?.note ?? '',
         product: product,
-        fixedInputs: subscription.fixedInputs,
-        customer: customer,
```
acidjunk, line 214: No customer in a subscription?

I though the backend would handle at least a name + customer ID
(fixed/hardcoded when not needed)

libs/orchestrator-ui-components/src/lib/utils/date.ts
```diff
@@ -7,7 +7,7 @@ export const parseDate = (date: string | null | undefined): Date | null => {
     }
 
     // Todo #95: GraphQL Dates will soon be in ISO8601 format instead of a float in a string
```
acidjunk, line 9: Is the todo still valid?

ricardovdheijden: It is partially valid still, since the harcoded "dutch locale" mentioned in bug 95, not tackled yet

acidjunk: Yup but that will probably be tackled in Ruben localisation ticket.
Better to remove the Todo: as dates are now indeed IS)8601 formatted

apps/example-orchestrator-ui/components/Subscription/subscriptionQuery.ts
ricardovdheijden: Currently the backend does not seem to support the customer prop

libs/orchestrator-ui-components/src/lib/utils/date.ts
ricardovdheijden: done

apps/example-orchestrator-ui/components/Subscription/subscriptionQuery.ts
ricardovdheijden: Added reference to the story as a todo that will add the prop back

acidjunk (approved):

PR #161, opened 2023-07-21
147 turbo repo migration

acidjunk:
- [ ] Merge branches into nx repo
- [ ] Copy all component files + App files over the files in the Turbo repo (so you have the latest changes there (then lint in Turbo repo)
- [ ] Throw away all NX config and package.json as well (in NX repo)
- [ ] Rename lib and app folders to match Turborepo: think about lib folder structure because it is hierarchically in a different place in the tree (you should mimic that first)
- [ ] Copy all turbo config (from Turbo repo NX)
- [ ] Copy all files and config from Turbo repo to the NX repo


DutchBen (approved):

acidjunk (commented):
Yeah

PR #179, opened 2023-07-28
119: Setup translations and translate current content

DutchBen:


packages/orchestrator-ui-components/src/messages/getTranslationMessages.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    it('Returns en-US translation when en-US locale is requested', () => {
+        const translation = getTranslationMessages(Locale.enUS);
+        expect(translation).toEqual(enUS);
+    });
+
+    it('Returns en-US translation when no locale is requested', () => {
+        const translation = getTranslationMessages(undefined);
+        expect(translation).toEqual(enUS);
+    });
+    it('Returns en-US translation unknown locale  is requested', () => {
```
tjeerddie, line 21:
```suggestion
    it('Returns en-US translation unknown locale is requested', () => {
```

apps/wfo-ui/pages/_app.tsx
```diff
@@ -56,35 +59,37 @@ function CustomApp({ Component, pageProps }: AppProps) {
...
-                <OrchestratorConfigProvider
-                    initialOrchestratorConfig={initialOrchestratorConfig}
-                >
-                    <QueryClientProvider
-                        client={queryClient}
-                        contextSharing={true}
+        <TranslationsProvider>
+            <EuiProvider colorMode="light" modify={defaultOrchestratorTheme}>
+                <Head>
+                    <title>Welcome to example-orchestrator-ui!</title>
```
tjeerddie, line 65: use translation

DutchBen: The translation context is not available here yet. This requires a rearrangement of the the app page that I think is better saved for a separate PR once the translation approach is established. Or shall I just do it here right away?

.husky/pre-commit
ricardovdheijden, line 1: What happended with this file? I only see a size diff.

apps/wfo-ui/pages/_app.tsx
ricardovdheijden: This _app is getting bigger and bigger with all the providers and contexts. Most of them should/might end up in the library to be consumed here. I would say to do fix this at that moment. I do agree that we eventually want to translate the title too.

apps/wfo-ui/translations/README.md
```diff
@@ -0,0 +1,44 @@
+## Adding or overriding translations
```
ricardovdheijden, line 1: Since the translations are part of the library as well, we could also add/move it there

ricardovdheijden (approved):

.husky/pre-commit
DutchBen: It was changed to be executable so ```chomd +x``` It wouldn't pick run the hook for me without this

DutchBen:
Looks like this PR is messed up after a rebase. Closing and reopening

PR #180, opened 2023-07-31
110 updates eslint package

ricardovdheijden:
The changes made in this PR should make it possible to just copy the app and run it standalone.
- Eslint was missing in the eslint package, moved it from the root-package.json to the config package
- Running stanalone requires the transpile prop to be populated in the nextjs config

### DIFF
--- a/packages/eslint-config-custom/package.json
+++ b/packages/eslint-config-custom/package.json
@@ -9,6 +9,7 @@
     },
     "dependencies": {
         "@typescript-eslint/eslint-plugin": "5.61.0",
+        "eslint": "^8.45.0",
         "eslint-config-next": "^13.4.1",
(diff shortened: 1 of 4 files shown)


packages/eslint-config-custom/package.json
```diff
@@ -9,6 +9,7 @@
     },
     "dependencies": {
         "@typescript-eslint/eslint-plugin": "5.61.0",
+        "eslint": "^8.45.0",
```
tjeerddie, line 12: eslint dependency instead of devdependency?

ricardovdheijden: yeah, and i think all of the others too. Will fix it.

DutchBen (approved):

tjeerddie (approved):

PR #189, opened 2023-08-02
86 metadata productblocks page

DutchBen:
### Summary
Creates the metadata product blocks page and move the page from the app into the package. 

Contains changes to the metadata product page that are part of one of these tickets but are added here to make the commit hoooks pass:
Productspage: https://github.com/workfloworchestrator/orchestrator-ui/issues/150
Fixed inputs: https://github.com/workfloworchestrator/orchestrator-ui/issues/88
### DOD Tasks

- NA: Add/Update documentation.
- NA: Update CHANGELOG.md.
- NA: Affected functionality covered by system tests.

### DIFF
--- a/apps/wfo-ui/components/Metadata/Products/Products.tsx
+++ b/apps/wfo-ui/components/Metadata/Products/Products.tsx
@@ -9,7 +9,6 @@ import {
     WFOProductBlockBadge,
     useQueryWithGraphql,
     getTypedFieldFromObject,
-    parseDateToLocaleString,
 } from '@orchestrator-ui/orchestrator-ui-components';
 
 import type {
@@ -30,7 +29,6 @@ import { useTranslations } from 'next-intl';
 import { GET_PRODUCTS_GRAPHQL_QUERY } from './productsQuery';
 
 import { METADATA_PRODUCT_TABLE_LOCAL_STORAGE_KEY } from '../../../constants';
-import { MetadataProductsQuery } from '../../../__generated__/graphql';
 import { useRouter } from 'next/router';
 import { mapToGraphQlSortBy } from '../../../utils/queryVarsMappers';
 
@@ -181,16 +179,3 @@ export const Products: FC<ProductsProps> = ({
         />
     );
 };
-/*
-function mapApiResponseToProductTableData(
-    graphqlResponse: MetadataProductsQuery,
-): ProductDefinition[] {
-    return graphqlResponse.products.page.map(
-        (product): ProductDefinition => ({
-            ...product,
-            createdAt: product.createdAt,
-            productId: 'TEMPID'
-        }),
-    );
-}
-*/
--- a/apps/wfo-ui/components/Metadata/Products/Products.tsx
+++ b/apps/wfo-ui/components/Metadata/Products/Products.tsx
@@ -9,13 +9,12 @@ import {
     WFOProductBlockBadge,
     useQueryWithGraphql,
     getTypedFieldFromObject,
-    parseDate,
     parseDateToLocaleString,
 } from '@orchestrator-ui/orchestrator-ui-components';
 
 import type {
     DataDisplayParams,
-    Product,
+    ProductDefinition,
 } from '@orchestrator-ui/orchestrator-ui-components';
 
 import {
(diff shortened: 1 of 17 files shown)


packages/orchestrator-ui-components/src/pages/metadata/WFOProductBlocksPage.tsx
```diff
@@ -0,0 +1,167 @@
...
+            width: '90',
+            render: (value) => (
+                <WFOStatusBadge status={value.toLocaleLowerCase()} />
+            ),
+        },
+        resourceTypes: {
+            field: PRODUCT_BLOCK_FIELD_RESOURCE_TYPES,
+            name: t('resourceTypes'),
+            render: (resourceTypes) => (
+                <>
```
ricardovdheijden, line 98: Fragment is not needed here

packages/orchestrator-ui-components/src/hooks/useDataDisplayParams.ts
```diff
@@ -12,7 +12,7 @@ import { DEFAULT_PAGE_SIZE } from '../components';
 export type DataDisplayParams<Type> = {
     pageSize: number;
     pageIndex: number;
-    sortBy?: GraphQLSort<Type>;
+    sortBy: GraphQLSort<Type>;
```
ricardovdheijden, line 15: I made set this to a possible undefined value since the data can be entered trough the query paramter and can potentially be missing or invalid. In case of missing or invalid it is undefined in the app. Or do you have some fallback mechanism while processing the parameters?

apps/wfo-ui/utils/queryVarsMappers.ts
```diff
@@ -14,9 +14,9 @@ export const mapSortOrderToGraphQlSortOrder = (
 
 export const mapToGraphQlSortBy = <Type>(
     sortBy: GraphQLSort<Type> | undefined,
-): GraphqlSort | null => {
+): GraphqlSort | undefined => {
```
ricardovdheijden, line 17: If sortBy in packages/orchestrator-ui-components/src/hooks/useDataDisplayParams.ts is always defined, we might want to update this function as well. Looks like the if statement might not be needed, since sortBy, sortBy.order and sortBy.field will not be undefined.

apps/wfo-ui/pages/metadata/productblocks.tsx
```diff
@@ -1,12 +1,6 @@
 import React from 'react';
-import { MetaDataLayout } from '../../components/Metadata/PageLayout/layout';
+import { WFOProductBlocksPage as Page } from '@orchestrator-ui/orchestrator-ui-components';
```
ricardovdheijden, line 2: A cast to "Page" does not seem to be needed here, since the router searches for the default export in this file

apps/wfo-ui/components/Metadata/Products/Products.tsx
```diff
@@ -99,20 +106,33 @@ export const Products: FC<ProductsProps> = ({
                 </>
             ),
         },
+        fixedInputs: {
+            field: PRODUCT_FIELD_FIXED_INPUTS,
+            name: t('fixedInputs'),
+            render: (fixedInputs) => (
+                <>
```
ricardovdheijden, line 113: fragment is not needed here

apps/wfo-ui/components/Metadata/Products/Products.tsx
```diff
@@ -99,20 +106,33 @@ export const Products: FC<ProductsProps> = ({
                 </>
             ),
         },
+        fixedInputs: {
+            field: PRODUCT_FIELD_FIXED_INPUTS,
+            name: t('fixedInputs'),
+            render: (fixedInputs) => (
```
ricardovdheijden, line 112: minor could: this whole block could be a FixedInputs component to end up wit a more compact tableColumns object here

packages/orchestrator-ui-components/src/pages/metadata/WFOProductBlocksPage.tsx
DutchBen: I think it is. The resourcetypes are mapped so you get multiple elements on that need to have something around them

packages/orchestrator-ui-components/src/hooks/useDataDisplayParams.ts
DutchBen: I give a default sorting with every page. That way it will always get a value even if it's not provided. Is that what you mean or is it something else? If it's optional I get errors in other places although I don't remember exactly where

apps/wfo-ui/pages/metadata/productblocks.tsx
DutchBen: Fixed

apps/wfo-ui/utils/queryVarsMappers.ts
DutchBen: Yes. Removed the undefined and the if

ricardovdheijden (approved):

PR #195, opened 2023-08-04
Updates tsconfig to define a root and out dir

ricardovdheijden:
This should make the IDE find the non-transpiled files when consuming the ui package in the app in the monorepo


tjeerddie (approved):
LGTM

DutchBen (commented):
Nice

PR #204, opened 2023-08-07
150 metadata products rewrite

DutchBen:


DutchBen:
Just tried to see if it would work on a newew branch

PR #215, opened 2023-08-11
91 - Process list page

ricardovdheijden:
Implements the ProcessList page.

Also includes:
- Refactor of the FilterTab component to support other pages than Subscription page
- Added feature to the table component to refresh the hidden columns when the key changes

PR does not include:
- Sorting: backend change is needed to match the colum names with the data (for now hardcoded)
- Missing columns: target, tag, customer abbrev -- waiting for backend change to add these
- lastStep prop: is implemented in frontend, but backend returns always 'null' as value


packages/orchestrator-ui-components/src/components/WFOFilterTabs/WFOFilterTabs.tsx
```diff
@@ -0,0 +1,46 @@
...
+export type WFOFilterTabsProps<TabType, DataType> = {
+    tabs: WFOFilterTab<TabType, DataType>[];
+    selectedSubscriptionsTab: TabType;
+    translationNamespace: string;
+    onChangeSubscriptionsTab: (updatedSubscriptionsTab: TabType) => void;
+};
+
+export const WFOFilterTabs = <TabType extends string, DataType>({
+    tabs,
+    selectedSubscriptionsTab,
```
tjeerddie, line 25: since this is named WFOFilterTabs, does it still need `Subscriptions` specifically in the props?

ricardovdheijden: Done

tjeerddie (approved):

Issue #188, opened 2023-08-01
Breadcrumb link of Metadata navigates to wrong page


wouter1975:
The link of Metadata navigates to a different page than the Metadata link of the left menu. It should load the Product page.

acidjunk:
Closed via MR

PR #222, opened 2023-08-14
Fix breadcrumb link of metadata

Georgi2704:
#188 

I tried to make the redirect programatically from the WFOBreadcrumbs component with router.push() or router.replace() so the page doesn’t reload unnecessary, this didn’t work so I found that I can do that from next.config.js. The minus is that it triggers a refresh when going back down one level in the breadcrumb component.

Also moved the paths to a separate file, but I can always put them back.

### DIFF
--- a/apps/wfo-ui/middleware.ts
+++ b/apps/wfo-ui/middleware.ts
@@ -2,7 +2,7 @@ import { NextResponse } from 'next/server';
 import type { NextRequest } from 'next/server';
 
 export function middleware(request: NextRequest) {
-    return NextResponse.rewrite(new URL('/metadata/products', request.url));
+    return NextResponse.redirect(new URL('/metadata/products', request.url));
 }
 
 // Supports both a single string value or an array of matchers


ricardovdheijden:
I might have found a solution for the reloading:
image
Create index.tsx in the page-router (apps/wfo-ui/pages/metadata/index.tsx) and make all links just route to '/metadata'

apps/wfo-ui/next.config.js
```diff
@@ -7,4 +7,13 @@ module.exports = {
         defaultLocale: 'en-US',
     },
     transpilePackages: ['@orchestrator-ui/orchestrator-ui-components'],
+    async redirects() {
```
ricardovdheijden, line 10: If you implement my suggested solution, this config is not needed anymore

acidjunk: Fixed the redirect with middleware: no need for a dummy page.

.changeset/fresh-pets-eat.md
```diff
@@ -0,0 +1,6 @@
+---
+'@orchestrator-ui/orchestrator-ui-components': patch
+'wfo-ui': patch
```
ricardovdheijden, line 3: WFO UI does not need a version bump.

acidjunk: Removed it.

I think it would be nice to have increasing version numbers (even when we don't publish a package to NPM)

That can also be done; when releasing it from main

ricardovdheijden (approved):

PR #240, opened 2023-08-17
91 - Renames the prop id to processId due to backend change

ricardovdheijden:


packages/orchestrator-ui-components/src/messages/nl-NL.json
```diff
@@ -62,7 +62,7 @@
             "subscriptions": "Subscriptions",
             "createdBy": "Gecreëerd door",
             "assignee": "Assignee",
-            "id": "ID",
+            "processId": "ID",
```
acidjunk, line 65: PID might be a better translation.

PR #250, opened 2023-08-22
Forms POC 1 - get working forms

acidjunk:
This is a POC implementation of the uniforms form part from the V1 GUI.

- [x] copy all the uniforms files and keep them together (in v1 they were scattered all over the place)
- [x] try to keep the form functionality in separate package
- [x] rewrite UserInputForm so it works in a Next JS context
- [x] install deps and come up with a strategy for the depencies
- [x] install next-intl : and refactor some Form Fields that use it
- [x] get one form working with CreateForm (CIM forms should be easier than a complete WF form flow)
- [x] implemented some basic form fields to know the amount of work
- [x] implemented axios client from v1
- [x] added `turbo reset`: which removes all node_modules
- [x] working FormValidation
- [x] Working ProductField
- [x] Working New Process Form (including preselects) 
- [ ] linting and typescript errors (~65 errors/warnings left)

Bug? validation when navigating back to form 1 of the wizard fails.


apps/wfo-ui/package.json
```diff
@@ -8,7 +8,8 @@
         "tsc": "tsc --noEmit",
         "build": "next build",
         "start": "next start",
-        "lint": "next lint"
+        "lint": "next lint",
+        "reset": "rm -rf node_modules"
```
ricardovdheijden, line 12: A reset script is not really needed. What would be the usecase?

packages/orchestrator-ui-components/src/components/WFOForms/AutoFieldLoader.tsx
```diff
@@ -0,0 +1,54 @@
...
+export function autoFieldFunction(
+    props: GuaranteedProps<unknown> & Record<string, any>,
+    uniforms: Context<unknown>,
+) {
+    const { fieldType, field } = props;
+    const { format } = field;
+
+    switch (fieldType) {
+        case Number:
+            switch (format) {
```
ricardovdheijden, line 21: a switch in a switch migt be better to be moved to a separate method.
case Number: getNumberFormat(format) -- containing the other switch

packages/orchestrator-ui-components/src/api/axios.ts
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import axios, { AxiosRequestConfig } from 'axios';
+
+const apiPath = 'http://localhost:8080/api/surf/forms';
```
ricardovdheijden, line 18: Having a localhost url in the code base is not preferred.
Is the form also available on the Dev environment? Or maybe in an Okteto namespace?

packages/orchestrator-ui-components/src/components/WFOForms/CreateForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+import React, { useCallback, useEffect, useState } from 'react';
+import { Form, FormNotCompleteResponse } from '../../types/forms';
+import UserInputFormWizard from './UserInputFormWizard';
+import { apiClient } from '../../api';
+
+interface IProps {
+    /* eslint-disable  @typescript-eslint/no-explicit-any */
+    preselectedInput?: any;
```
ricardovdheijden, line 23: Could you specify why the any is needed here? maybe try to fall back on "string" or "object of key values"

packages/orchestrator-ui-components/src/components/WFOForms/UserInputFormStyling.ts
```diff
@@ -0,0 +1,89 @@
+// Todo: theme with decent colors from theme (style was copied from v1)
+const PRIMARY_COLOR = '#0077cc';
+const LIGHT_PRIMARY_COLOR = '#520980';
+
+import { css } from '@emotion/react';
+
+export const userInputFormStyling = css`
```
ricardovdheijden, line 7: A few generic comments about styling:
- In the rest of the library we are using the css object way of specifying the styles, that also solves getting the theme into the styles.
- When a component has multiple files (for example styles and later storybook) you might want to move each component in its own folder -- at least the ones containing multiple files

apps/wfo-ui/package.json
acidjunk: Do you know a simple way to remove all node_modules folders from all packages? So you can start clean (e.g. probably handy when we upgrade react or Next)

packages/orchestrator-ui-components/src/components/WFOForms/AutoFieldLoader.tsx
acidjunk: Copied from V1; the actual swithc will be a lot bigger when we add all Fields

packages/orchestrator-ui-components/src/api/axios.ts
acidjunk: It's merged now on dev.

Will adapt in #274 : which is the branch that will probably be merged.

packages/orchestrator-ui-components/src/components/WFOForms/CreateForm.tsx
acidjunk: No not yet.

The preselectInput thing can probably be removed; as there is no use case yet for having preselectedInput in createForm.

When the form is applied in a process context; this needs to be addressed.

DutchBen:
Closed. Implemented in MVP-Process detail/Note-modify branch

Issue #211, opened 2023-08-09
Create table row view modal


wouter1975:
Add a modal to view all records of a row of the table component. Some of the cells contain more information which is not visible in the table.

**Requirements:**
- [x] Add a view icon in each row, it should open the modal
- [x] use Table context in the header of the modal (eg `Details Product Block`)
- [x] The modal shall provide information of all the columns of the table, including the hidden columns.
- [x] Every `boxed` value like resource type (blue shaded) should be written on a new line
- [x] Use the same formatting like in the table
- [x] Add close button
- [x] ~~Modal can be resized~~ Modal should have overflow
- [x] Add a copy action to each row --> on hover
- [x] Sort the keys in the detail view the same as the table itself (iterate the tableConfig object instead of the data)
- [x] Implement in all pages/tables
- [x] Render data the same way as in table , but always show all (in case of multiple badges)
- [ ] Make multiple badges render nicely (with top/bottom margin)
- [x] Show UUID's in the shorter form in the table and fully in the detail view (all tables in project)

Design could look like:

![Image](https://user-images.githubusercontent.com/83226590/259456542-fdd51d51-f277-409d-a20b-4e8a318574d9.png)

Note that Vishal already prepared such a detail view design voor the IMS details. Please follow his design as much as possible.

![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/a868d96d-cf7b-457f-9a47-ce2130f9af12)

PR #253, opened 2023-08-24
211 detail view match key order table

ricardovdheijden:
#211 

- Fixes the unsorted key order in the detail view (in table component)
- Shortens UUIDs in tables
- Improves rendering multiple badges per cell in the detail view

### DIFF
--- a/packages/orchestrator-ui-components/src/pages/metadata/WFOProductBlocksPage.tsx
+++ b/packages/orchestrator-ui-components/src/pages/metadata/WFOProductBlocksPage.tsx
@@ -31,6 +31,7 @@ import { useDataDisplayParams, useQueryWithGraphql } from '../../hooks';
 import { GET_PRODUCTS_BLOCKS_GRAPHQL_QUERY } from '../../graphqlQueries';
 
 import { WFOMetadataPageLayout } from './WFOMetadataPageLayout';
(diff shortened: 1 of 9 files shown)


packages/orchestrator-ui-components/src/pages/metadata/WFOProductBlocksPage.tsx
```diff
@@ -70,7 +68,9 @@ export const WFOProductBlocksPage = () => {
             field: PRODUCT_BLOCK_FIELD_ID,
             name: t('id'),
             width: '90',
-            render: (value) => getFirstUuidPart(value),
+            render: (value) => (
+                <span title={value}>{getFirstUuidPart(value)}</span>
+            ),
             renderDetails: (value) => value,
         },
```
mrijk, line 75: Maybe it's possible to extract some common code for this part (render a short UUID). It is now in at least 4 places

mrijk (approved):
I think we could avoid some code duplication

packages/orchestrator-ui-components/src/pages/metadata/WFOProductBlocksPage.tsx
ricardovdheijden: good one, introduced a component for rendering a short uuid

PR #263, opened 2023-08-31
218 multiple tsconfigs

ricardovdheijden:
This PR introduces a specific build tsconfig in the orchestrator. It resolves the go-to issue when clicking from the app on a component to go right to the source code in the package instead of its dist folder.
While running the dev-server it still depends on the dist folders, which is desired.

Building the packages in the prepare script is need to prevent welcoming the developer with all kinds of not-found errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
         "build": "turbo run build",
         "packages:changeset": "changeset",
         "packages:publish": "turbo run build lint test && changeset publish",
-        "prepare": "husky install"
+        "prepare": "husky install && turbo run build --filter=./packages/*"
     },
     "devDependencies": {
         "@changesets/cli": "^2.26.2",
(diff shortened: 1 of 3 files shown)


packages/orchestrator-ui-components/tsconfig.json
```diff
@@ -1,9 +1,11 @@
 {
     "extends": "@orchestrator-ui/tsconfig/base.json",
     "compilerOptions": {
-        "strictNullChecks": true
+        "strictNullChecks": true,
+        "rootDir": "./src",
```
acidjunk, line 5: Do you think we should also apply this to the other packages?

ricardovdheijden: It should only apply to packages containing typescript. Even the package @orchestrator-ui/tsconfig is not needed, as it only exports json files.

DutchBen (approved):

Issue #151, opened 2023-07-12
Move components down to lib for Subscriptions detail page


DutchBen:
Move components down to the lib until we can import the entire Subscription detail page from the lib
- all props are optional
- apply default values
- expose defaults

PR #264, opened 2023-09-01
151 move subscription detail to package

ricardovdheijden:
#151 
Moves the subscription detail and it page to the package.

Extra:
- Removes generated types and adds types in the package as replacement


packages/orchestrator-ui-components/src/components/WFOSubscription/WFOSubscription.tsx
```diff
@@ -59,17 +57,25 @@ export const Subscription: FC<SubscriptionProps> = ({ subscriptionId }) => {
     //         ).node,
     // };
 
+    const graphQLClient = new GraphQLClient(graphqlEndpointCore);
+
     const fetchSubscriptionOutline = async () => {
         console.log('Fetch outline query results for ID: ', subscriptionId);
```
mrijk, line 63: console.log is a bit evil, isn't it.. I vaguely remember from the past that I used some library for that so that you can easily toggle this.

packages/orchestrator-ui-components/src/pages/subscriptions/WFOSubscriptionDetailPage.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const WFOSubscriptionDetailPage = () => {
+    const router = useRouter();
+    const { subscriptionId } = router.query;
+
+    return (
+        <SubscriptionProvider>
+            <TreeProvider>
+                {subscriptionId && (
+                    <WFOSubscription
+                        subscriptionId={
```
mrijk, line 15: Without knowing the details, this kind of code (checking on array) breaks of course every typing system, but is also a bit vague. So subcriptionId can be either a scalar or an array?

packages/orchestrator-ui-components/src/types/subscription.ts
```diff
@@ -12,14 +12,32 @@ export enum SubscriptionStatus {
 export type Subscription = {
     subscriptionId: string;
     description: string;
-    note: string;
+    note?: string;
     startDate: string;
```
mrijk, line 16: Maybe we discussed this before (I forgot), but why is a date represented as a string and not a type?

packages/orchestrator-ui-components/src/components/WFOSubscription/WFOSubscription.tsx
ricardovdheijden: Yes indeed it doesn't belong there. For now I kept de scope to only move to lib. A different story is finalizing the component itself.

tjeerddie (approved):

packages/orchestrator-ui-components/src/types/subscription.ts
ricardovdheijden: The actual type of the field being returned from the API is a string. It is shaped as a date (ISO). I do admit that we might want to move to 2 different types of "Subscription": one that is the actual shape of the API (startDate is string) and one that represents where the data actually stands for (startDate is a Date, id is a UUID etc). Currently we parse the date very close to the component itself

packages/orchestrator-ui-components/src/pages/subscriptions/WFOSubscriptionDetailPage.tsx
ricardovdheijden: True, but unfortunately the type of subscriptionId is actually any of the 2 types: 
image

mrijk: Ok, wouldn't then make sense to isolate this 'ugliness' and just create a function that always returns a string | undefined? And can it indeed be undefined?

PR #274, opened 2023-09-08
Form POC 3 - implement form api with react query

acidjunk:
This is a POC implementation of the uniforms form part from the V1 GUI.

- [x] copy all the uniforms files and keep them together (in v1 they were scattered all over the place)
- [x] try to keep the form functionality in separate package
- [x] rewrite UserInputForm so it works in a Next JS context
- [x] install deps and come up with a strategy for the depencies
- [x] install next-intl : and refactor some Form Fields that use it
- [x] get one form working with CreateForm (CIM forms should be easier than a complete WF form flow)
- [x] implemented some basic form fields to know the amount of work
- [x] implemented axios client from v1
- [x] added `turbo reset`: which removes all node_modules
- [x] added react-query mutation implementation to start the form
- [x] fixed and tested nullable fields 
- [x] handle subsequent submits (wizard) also with fetch
- [x] implement SelectField so #291 is possible
- [x] removed axios
- [ ] change form endpoint to ochestrator dev (which now contains example form)
- [ ] linting and typescript errors (~65 errors/warnings left)

Researched the ways to fetch data in Next JS. The API Routes / [Route handlers](https://nextjs.org/docs/app/building-your-application/routing/route-handlers) seem to be limited to routes; e.g. tied to a specific Next JS page; that will them delegate the API call to the next JS server. Primary use case: page level cache and to hide stuff like an API key that the backend uses to query stuff from another micro service.

For client side only the docs suggest: react-query or SWR; an almost similar project powered by the same team that's behind Next JS. On the surface, it seems to only lack a global accesible cache.

In the future there seem to be plans to adopt: `use` -> [info](https://github.com/acdlite/rfcs/blob/first-class-promises/text/0000-first-class-support-for-promises.md#usepromise)
But the docs also mention ways to handle form data in Next JS with server sided handling; via [server action](https://nextjs.org/docs/app/building-your-application/data-fetching/forms-and-mutations) : this seems tailored on form handling; but I don't see an easy way to do our uniforms data flow with it.

It would be nice to drop the `axios` dependency and try to get it working with react-query for now. So focused on trying to  get it working with react-query and fetch; most of the other things use fetch under the hood, so it should be easy to switch when needed.

**Note**: It might be worthwhile, as @DutchBen mentioned, to investigate other form libs as well; as Next JS seems to have quite some form batteries included. A nice comparison can be found [here](https://uniforms.tools/docs/compare-matrix/)

Bug? validation when navigating back to form 1 of the wizard fails.


DutchBen:
Implemented in MVP-Process detail/Note-modify branch

Issue #223, opened 2023-08-15
Process detail page: implement time line component


wouter1975:
Implement the time line component

![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/158f9744-3bed-42d4-a469-654cd2b1e6b4)

Add process time line component, this component has the attributes: 
* total number of steps
* current step index
* step description

In the first implementation add the following features:
- [x] visualize the current step as part of time line
- [x] use full width of component, and equally spaced `dots` to represent a process step
- [x] steps < step_index should have the completed style
- [x] steps = step_index should have the active style
- [x] steps > step_index should have the disabled style
- [x] add mouse over effect on each `dot`, displaying the step description
  - [ ] this mouse over can be enhanced with more information at a later stage

PR #275, opened 2023-09-10
223 process detail time line

ricardovdheijden:
#223 , Rendering a timeline for the steps in a process.

In case a step is retried or in some way executed twice, both steps get visually combined with a '2' in the timeline:
image


packages/orchestrator-ui-components/src/components/WFOTimeline/WFOTimeline.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const mapTimelineItemToEuiStep = ({
+        processStepStatus,
+        value,
+        onClick,
+    }: TimelineItem): EuiStepHorizontalProps => {
+        const euiStepStatus =
+            mapProcessStepStatusToEuiStepStatus(processStepStatus);
+
+        return {
+            status: typeof value === 'number' ? undefined : euiStepStatus,
```
DutchBen, line 67: What does this do? It returns undefined if the status is a number but when does this situation occur?  It feels this could be handled in a different place

packages/orchestrator-ui-components/src/components/WFOTimeline/mapProcessStepStatusToEuiStepStatus.ts
```diff
@@ -0,0 +1,22 @@
...
+import { EuiStepStatus } from '@elastic/eui/src/components/steps/step_number';
+
+export const mapProcessStepStatusToEuiStepStatus = (
+    processStepStatus: StepStatus,
+): EuiStepStatus => {
+    switch (processStepStatus) {
+        case StepStatus.SUCCESS:
+        case StepStatus.SKIPPED:
+        case StepStatus.COMPLETE:
+            return 'complete';
```
DutchBen, line 11: Does Eui expose an enum for the step statusses that you can use for the return value?

packages/orchestrator-ui-components/src/pages/processes/timelineUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+            console.log(`Clicked on ${step.name} (${step.stepid})`);
+        },
+    };
+};
+
+const stepsShouldBeMerged = (
+    previousStep: ProcessDetailStep,
+    currentStep: ProcessDetailStep,
+) =>
+    previousStep.status !== StepStatus.COMPLETE &&
```
DutchBen, line 27: I would like to see a test for this logic. I think StepStatus.SUCCES should also be checked for the edge case you have 2 steps with the same name in succession.

ricardovdheijden: Good one, an probably SKIPPED as well

packages/orchestrator-ui-components/src/components/WFOTimeline/WFOTimeline.tsx
ricardovdheijden: I made the results to be returned a bit more explicit (3 cases)

packages/orchestrator-ui-components/src/pages/processes/timelineUtils.ts
ricardovdheijden: Introduced a helper function to determine if a status is considered final

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WFOTimeline/styles.ts
```diff
@@ -0,0 +1,154 @@
...
+        display: 'flex',
+        flexDirection: 'row',
+        alignItems: 'center',
+    });
+
+    return {
+        timelinePanelStyle,
+        stepStyle,
+        getStepLineStyle,
+        stepOuterCircleStyle: getStepOuterCircleStyle,
```
DutchBen, line 151: why the rename?

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WFOTimeline/styles.ts
ricardovdheijden: It used to be plain style before , but now its a function. Refactor in the IDE left the previous name to maintain the interface for the consuming components. -- Not needed, fixed it now

PR #292, opened 2023-09-14
91 process list page api update

ricardovdheijden:
Processes changes in types due to backend changes. The scope of this PR is only ProcessList and TaskList

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WFOProcessesList/processListObjectMappers.ts
+++ b/packages/orchestrator-ui-components/src/components/WFOProcessesList/processListObjectMappers.ts
@@ -47,17 +47,6 @@ export const mapGraphQlProcessListResultToProcessListItems = (
 // Backend concatenates object name with the key, e.g. product.name becomes productName
 const fieldMapper = (field: keyof ProcessListItem): keyof Process => {
     switch (field) {
-        case 'workflowName':
-        case 'lastStep':
-        case 'lastStatus':
-        case 'workflowTarget':
-        case 'createdBy':
-        case 'assignee':
-        case 'processId':
-        case 'startedAt':
-        case 'lastModifiedAt':
-        case 'subscriptions':
-            return field;
         case 'customer':
             return 'customerFullname' as keyof Process;
         case 'customerAbbreviation':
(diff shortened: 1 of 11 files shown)


packages/orchestrator-ui-components/src/components/WFOProcessesList/processListObjectMappers.ts
```diff
@@ -0,0 +1,79 @@
...
+            lastModifiedAt,
+            subscriptions,
+            productName: product?.name,
+            productTag: product?.tag,
+            customer: customer.fullname,
+            customerAbbreviation: customer.shortcode,
+        };
+    });
+
+// Some fields are not a key of Process, however backend still supports them
```
DutchBen, line 46: This explanation raises some questions to me. Isn't the case that some fields in the process list are not a key of a Process but a key of a nested type of a key of Process? Or is that what you are saying?

packages/orchestrator-ui-components/src/components/WFOProcessesList/processListObjectMappers.ts
```diff
@@ -0,0 +1,79 @@
...
+// Backend concatenates object name with the key, e.g. product.name becomes productName
+// Todo: typecast is needed until ticket is implemented:
+// https://github.com/workfloworchestrator/orchestrator-ui/issues/290
+const fieldMapper = (field: keyof ProcessListItem): keyof Process => {
+    switch (field) {
+        case 'customer':
+            return 'customerFullname' as keyof Process;
```
DutchBen, line 53: Is it an option to return customer.fullName here? That makes it a little clearer to me. Or wont that work?

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WFOProcessesList/processListObjectMappers.ts
ricardovdheijden: Not exactly. It is for sure "key of Process" but additonally some concatenation of the object name and field.
`product: { name: string }` -- becomes productName. Though backend did not applied any types there. Backend assumes the field is a string instead of "keyof Process".

PR #293, opened 2023-09-14
225 StepsCards in processDetail 

DutchBen:


packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -0,0 +1,129 @@
...
+                                toggleStepDetailIsOpen={toggleStepDetailIsOpen}
+                                step={step}
+                                stepIndex={index}
+                                startedAt={startedAt}
+                            />
+                        </div>
+                    );
+
+                    if (index > 0) {
+                        startedAt = step.executed;
```
ricardovdheijden, line 122: It looks like you are overriding an input prop of the steplist component here. You could do this inline without this reassignment (see comment above)

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -0,0 +1,129 @@
...
+                        <div key={`step-${index}`}>
+                            {index !== 0 && <div css={stepSpacerStyle} />}
+                            <WFOStep
+                                stepDetailIsOpen={
+                                    stepDetailStates.get(index) || false
+                                }
+                                toggleStepDetailIsOpen={toggleStepDetailIsOpen}
+                                step={step}
+                                stepIndex={index}
+                                startedAt={startedAt}
```
ricardovdheijden, line 116: this could be `index > 0 ? step.executed : startedAt`

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
```diff
@@ -0,0 +1,91 @@
...
+                                <EuiText size="m">
+                                    {calculateTimeDifference(
+                                        startedAt,
+                                        step.executed,
+                                    )}
+                                </EuiText>
+                            </EuiFlexItem>
+                            <EuiFlexItem
+                                grow={0}
+                                css={{ marginRight: '8px', cursor: 'pointer' }}
```
ricardovdheijden, line 76: Try to use values from the theme. There is a theme.base value that can be used

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepStatusIcon/WFOStepStatusIcon.tsx
```diff
@@ -0,0 +1,73 @@
...
+    stepStatus: StepStatus;
+    theme: EuiThemeComputed;
+}
+
+const SubIcon = ({ stepStatus, theme }: SubIconProps) => {
+    switch (stepStatus) {
+        case StepStatus.PENDING:
+            return <></>;
+        case StepStatus.SUSPEND:
+            return <WFOMinusCircleFill color={'#8E6A00'} />;
```
ricardovdheijden, line 25: isnt this the colors.warningText? (same goes for different colors at line 46, 48 and 50 )

packages/orchestrator-ui-components/src/components/WFOSteps/getStyles.ts
```diff
@@ -0,0 +1,81 @@
+import { EuiThemeComputed } from '@elastic/eui/src/services/theme/types';
+import { css } from '@emotion/react';
+// import { makeHighContrastColor } from '@elastic/eui';
```
ricardovdheijden, line 3: commented code

packages/orchestrator-ui-components/src/components/WFOSteps/getStyles.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export const getStyles = (theme: EuiThemeComputed) => {
```
ricardovdheijden, line 5: Whole file: try to use data from the theme to populate the numeric values. You could use the theme.base value to multiply from there. There is also a multiplyBase() helper function in our theme, unless you pass it in here as a parameter

packages/orchestrator-ui-components/src/icons/WFOCogFill.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const WFOCogFill: FC<WFOIconProps> = ({
+    width = 24,
+    height = 24,
+    color = '#000000',
+}) => {
+    return (
+        <svg
+            width={width}
+            height={height}
+            viewBox={`0 0 ${width} ${height}`}
```
ricardovdheijden, line 13: Just as a warning to double check. In other icons you (and myself) did not make the viewBox configurable, just hardcoded 24 24. In my experience I got sometimes not-so-nice icons.

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
DutchBen: This doesn't do the same thing. The startedAt time is updated for every step so it get's the executed time from the previous step to be able to calculate the duration of the step

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
DutchBen: Changed it to theme.base/2

packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -217,11 +225,12 @@ export enum ProcessStatus {
     COMPLETED = 'completed',
 }
 
-export interface ProcessDetailStep {
+export interface Step {
```
ricardovdheijden, line 228: I would say that the previous name was better, as it is more descriptive, ProcessStep might also be nice

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepStatusIcon/WFOStepStatusIcon.tsx
DutchBen: Yeah so. Close a bit but not really. These colors are not in the theme. They are also quite specific to these icons so I thought it was ok to leave them here. I will ask Wouter what to do here, change the design color to standard warning or use a one of color

packages/orchestrator-ui-components/src/components/WFOSteps/getStyles.ts
DutchBen: Removed

packages/orchestrator-ui-components/src/components/WFOSteps/getStyles.ts
DutchBen: For the stepState icons the values are very tightly coupled. Changing the width will require a another value for the transform to get the icon to position correctly on top of the other and that might not be in de same proportions so I do want to make those fixed tbh.

packages/orchestrator-ui-components/src/types/types.ts
DutchBen: I can be a task step asweel thought right?

packages/orchestrator-ui-components/src/utils/date.ts
```diff
@@ -9,6 +9,35 @@ export const parseDate = (date: string | null | undefined): Date | null => {
     return new Date(date);
 };
 
+export const calculateTimeDifference = (from: string, to: string): string => {
+    const fromDate = parseDate(from);
+    const toDate = parseDate(to);
+
+    if (!fromDate || !toDate) return 'missing paramater';
```
ricardovdheijden, line 16: According to the parameter types this situation is not possible. Else the type would be string | undefined.
About the handling I would suggest to not return an error message as a string, but rather return an empty string or "0:00:00". Similar idea the "negative difference" return below

DutchBen: This supresses typescript errors in the code below that come from parseDate possibly returning Null so I do prefer leaving it in. And who nows, it might get triggered somewhere in the future somehow

I've considered returning 0.00.00 but that feels incorrect since it's not true. I've also considered throwing for both cases where I return a string now but that creates the need to handle the exception. This way at least you see the error very clearly in frontend code calling it. Toughts?

packages/orchestrator-ui-components/src/icons/WFOCogFill.tsx
DutchBen: This is needed to get the smaller icons from the design. It looks ok at least on my machine.

packages/orchestrator-ui-components/src/utils/date.ts
ricardovdheijden: parseDate possibly null, check! And my thought on returning values, I would go for empty string or "?". These return values seem more debug info. Maybe do check the requirements of these corner cases with Wouter/Hans

packages/orchestrator-ui-components/src/types/types.ts
ricardovdheijden: True, but tasks are still Processes and are typed like that

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
ricardovdheijden: Ok, but still reassigning a value from props is tricky. Is it possible to use useState here?

DutchBen: I think useState would cause a rerender everytime it's set. I have copied the startvalue to another variable so it doesn't change the prop anymore

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/utils/date.ts
DutchBen: ok. changed it to empty string

PR #294, opened 2023-09-15
Show state delta in steps

acidjunk:
Todo:

- [x] move state delta function to utils and try to remove the only `ts-ignore` in it
- [x] add a test for state delta function
- [x] translations
- [x] implement GUI stuff for failed steps
- [x] implement GUI stuff for input steps
- [x] clean up remaining inline style definitions in `WFOSteps` (and maybe also in `WFOStepList`?)

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
@@ -35,19 +35,18 @@ export const WFOStep = ({
 }: WFOStepProps) => {
     const { name, executed, status } = step;
     const { theme } = useOrchestratorTheme();
-    const { stepListContentBoldTextStyle, stepDurationStyle } =
-        getStyles(theme);
+    const {
+        stepContainerStyle,
+        stepListContentBoldTextStyle,
+        stepDurationStyle,
+        stepToggleExpandStyle,
+        stepRowStyle,
+    } = getStyles(theme);
     const t = useTranslations('processes.steps');
 
     return (
         <EuiPanel>
-            <EuiFlexGroup
-                css={{
-                    gap: 0,
-                    alignItems: 'center',
-                    marginBottom: 20,
-                }}
-            >
+            <EuiFlexGroup css={stepContainerStyle}>
                 <WFOStepStatusIcon stepStatus={status} />
 
                 <EuiFlexItem grow={0}>
@@ -57,13 +56,7 @@ export const WFOStep = ({
                     </EuiText>
                 </EuiFlexItem>
 
-                <EuiFlexGroup
-                    css={{
-                        flexGrow: 1,
-                        alignItems: 'center',
-                        justifyContent: 'flex-end',
-                    }}
-                >
+                <EuiFlexGroup css={stepRowStyle}>
                     {step.executed && (
                         <>
                             <EuiFlexItem
@@ -82,10 +75,7 @@ export const WFOStep = ({
                             </EuiFlexItem>
                             <EuiFlexItem
                                 grow={0}
-                                css={{
-                                    marginRight: theme.base / 2,
-                                    cursor: 'pointer',
-                                }}
+                                css={stepToggleExpandStyle}
                                 onClick={() =>
                                     toggleStepDetailIsOpen(stepIndex)
                                 }
--- a/packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
@@ -1,9 +1,15 @@
 import React from 'react';
-import { EuiFlexGroup, EuiFlexItem, EuiPanel, EuiText } from '@elastic/eui';
+import {
+    EuiCodeBlock,
+    EuiFlexGroup,
+    EuiFlexItem,
+    EuiPanel,
+    EuiText,
+} from '@elastic/eui';
 import { useTranslations } from 'next-intl';
 
 import { useOrchestratorTheme } from '../../../hooks';
-import type { Step } from '../../../types';
+import type { State, Step } from '../../../types';
 import { WFOStepStatusIcon } from '../WFOStepStatusIcon';
 import { getStyles } from '../getStyles';
 import { formatDate } from '../../../utils';
@@ -12,6 +18,7 @@ import { calculateTimeDifference } from '../../../utils';
 
 export interface WFOStepProps {
     step: Step;
+    delta: State;
     stepIndex: number;
     stepDetailIsOpen: boolean;
     toggleStepDetailIsOpen: (index: number) => void;
@@ -20,6 +27,7 @@ export interface WFOStepProps {
 
 export const WFOStep = ({
     step,
+    delta,
     stepDetailIsOpen,
     toggleStepDetailIsOpen,
     stepIndex,
@@ -37,6 +45,7 @@ export const WFOStep = ({
                 css={{
                     gap: 0,
                     alignItems: 'center',
+                    marginBottom: 20,
                 }}
             >
                 <WFOStepStatusIcon stepStatus={status} />
@@ -89,6 +98,16 @@ export const WFOStep = ({
                     )}
                 </EuiFlexGroup>
             </EuiFlexGroup>
+            {stepDetailIsOpen && (
+                <EuiCodeBlock
+                    isCopyable={true}
+                    language={'json'}
+                    lineNumbers={true}
+                    overflowHeight={6000}
+                >
+                    {JSON.stringify(delta, null, 3)}
+                </EuiCodeBlock>
+            )}
         </EuiPanel>
     );
 };
--- a/packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
@@ -36,7 +36,8 @@ export const WFOStep = ({
     const { name, executed, status } = step;
     const { theme } = useOrchestratorTheme();
     const {
-        stepContainerStyle,
+        stepHeaderStyle,
+        stepHeaderRightStyle,
         stepListContentBoldTextStyle,
         stepDurationStyle,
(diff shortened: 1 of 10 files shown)


packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -11,6 +11,33 @@ export interface WFOStepListProps {
     startedAt: string;
 }
 
+export const stateDelta = (prev: State, curr: State) => {
```
DutchBen, line 14: Matter of taste but I'm more for complete words instead of abbrv. So previous instead of prev

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -104,6 +131,32 @@ export const WFOStepList = ({ steps = [], startedAt }: WFOStepListProps) => {
             </EuiFlexGroup>
             <>
                 {steps.map((step, index) => {
+                    let prevState = {};
+                    let delta = {};
+
+                    if (index > 0) {
+                        // Todo: decided if we want this here.
+                        //  V1 has a rather big case statement with businness logic here
+                        if (steps[index].status === 'success') {
```
DutchBen, line 140: There is a enum with statusses somewhere

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -104,6 +131,32 @@ export const WFOStepList = ({ steps = [], startedAt }: WFOStepListProps) => {
...
+                            // Prepare a delta with the last successful step
+                            let prev_index = index - 1;
```
DutchBen, line 142: camelCase or snake_case? I see both used. I prefer camelCase but at least it should be consistent

DutchBen:
General comment. Is see there are todo's in the pr, do we still do a review if there are open todo's? What do you guys prefer?

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
```diff
@@ -89,6 +98,16 @@ export const WFOStep = ({
...
                 </EuiFlexGroup>
             </EuiFlexGroup>
+            {stepDetailIsOpen && (
+                <EuiCodeBlock
+                    isCopyable={true}
+                    language={'json'}
+                    lineNumbers={true}
+                    overflowHeight={6000}
+                >
+                    {JSON.stringify(delta, null, 3)}
```
ricardovdheijden, line 108: Maybe it is me, but I cannot stand a spacing of 3. Normally IDE's have either 2 or 4 configured as well. Personally I am team "4-spaces", but can live with 2 spaces as well.

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
```diff
@@ -37,6 +45,7 @@ export const WFOStep = ({
                 css={{
                     gap: 0,
                     alignItems: 'center',
+                    marginBottom: 20,
```
ricardovdheijden, line 48: Try to keep styling in a separate styles.ts file. More important, try to use the base value (theme.base) -- see packages/orchestrator-ui-components/src/components/WFOTimeline/styles.ts as example

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
ricardovdheijden: `step[index]` is the same as `step` in this case

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -104,6 +131,32 @@ export const WFOStepList = ({ steps = [], startedAt }: WFOStepListProps) => {
             </EuiFlexGroup>
             <>
                 {steps.map((step, index) => {
+                    let prevState = {};
```
ricardovdheijden, line 134: For me personally it is very hard to follow what is going on in this mapper, especially with the reassignments of prevState and delta. Try to move this logic as much as possible into a separate function, perferably in a separate ts file with unit tests.

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -104,6 +131,32 @@ export const WFOStepList = ({ steps = [], startedAt }: WFOStepListProps) => {
...
+
+                    if (index > 0) {
+                        // Todo: decided if we want this here.
+                        //  V1 has a rather big case statement with businness logic here
+                        if (steps[index].status === 'success') {
+                            // Prepare a delta with the last successful step
+                            let prev_index = index - 1;
+                            while (
+                                prev_index > 0 &&
+                                (steps[prev_index].status === 'failed' ||
```
ricardovdheijden, line 145: To make this function a bit more pure: you could use the 3rd argument from the mapper function that is the whole array too

packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -225,12 +225,16 @@ export enum ProcessStatus {
     COMPLETED = 'completed',
 }
 
+export interface State {
+    [index: string]: object;
```
ricardovdheijden, line 229: According to data on our test environment the type of the props can also be string, boolean, null (and maybe more)
image

packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -225,12 +225,16 @@ export enum ProcessStatus {
     COMPLETED = 'completed',
 }
 
+export interface State {
```
ricardovdheijden, line 228: Suggesting to name the type StepState or something more descriptive

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -11,6 +11,33 @@ export interface WFOStepListProps {
...
+
+    const newState = newKeys.sort().reduce((acc: State, key) => {
+        if (
+            curr[key] === Object(curr[key]) &&
+            !Array.isArray(curr[key]) &&
+            prevOrEmpty[key]
+        ) {
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
+            acc[key] = stateDelta(prevOrEmpty[key], curr[key]);
```
ricardovdheijden, line 32: Could you specify -- add as code comment -- why the TS ignore is needed here?

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
ricardovdheijden: Maybe even previousState (or if agree with my type name suggestion in a different comment previousStepState)

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
acidjunk: So why is the gap and alignItems ok here?

packages/orchestrator-ui-components/src/types/types.ts
acidjunk: Yes; I wonder why the step state was defined in a way that it's incompatible with the graphql response in the first place...

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
acidjunk: A mix was used in v1: and this was a copy, not yet ready for reviewing.

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStepList/WFOStepList.tsx
acidjunk: No; it's was already in the todo. I will when it's still needed after the mentioned refactor.

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
acidjunk: There are at least 6 other occurences of inline CSS. I will resolve them; no problem -> but you might review stuff better before it's merged. Added it to the todo list.

ricardovdheijden: Only since recent I came up with a pattern to move the inline css out of the components and having the theme available in that separate file. You will find many more places in the codebase where we still have inline CSS. So my comment about inline you can see as a "you could....". Though I would like to suggest to use the base value from the theme as a base to come up with these values

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
acidjunk: 4 won't fit with large workflows (e.g. fw)
2 looks flimsy

packages/orchestrator-ui-components/src/components/WFOSteps/WFOStep/WFOStep.tsx
acidjunk: @OFFTOPIC:
I now see a lot of css in getStyles.ts which hard codes color. This will get hard to maintain with a dark theme.
We need to come up with some good rules for colors.

packages/orchestrator-ui-components/src/types/types.ts
acidjunk: If you know a better solution please apply it in the MVP branch

DutchBen:
Already incorporated in MVP Branch

Issue #276, opened 2023-09-12
Add Dropdown menu when clicking New process button


wouter1975:
Use the design as leading design:

![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/8d30f845-14c3-4a06-bac9-f20787c0c7c6)

- [x] Open a modal with a dropdown component filled with all available product instances for which a create workflow exists. 
- [x] Clicking a product instance from the dropdown will launch the new process detail page an start the workflow for the selected product.
- [x] the modal should consist of a vertical scroll function
- [x] the modal should support a search input field, to filter the list of products

**Decision: let's rename new process to `New subscription`**

ricardovdheijden:
Wating for backend to support to filterBy "CREATE" target

wouter1975:
@ricardovdheijden Looks good! I see it is already merged.

In the dropdown the description can do without the Create... It makes it a bit more concise, as some of the descriptions are a bit long (maybe some cleanup needed in the product description as well).

PR #296, opened 2023-09-17
276 new process button combobox

ricardovdheijden:
#276 

Implementation of the new-subscription button with dropdown.
Added todo for waiting for backend support for the filtering on target, for now showing all workflows

After selecting a workflow, user gets redirected to a landing page (to be implemented in future story)
/new-subscription/WORKFLOW_NAME

image


apps/wfo-ui/pages/new-subscription/[processName].tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { useRouter } from 'next/router';
+
+const NewSubscriptionPage = () => {
```
Comment (DutchBen) [inline apps/wfo-ui/pages/new-subscription/[processName].tsx:4]: I think we are creating a new process here that doesn't necessarily has to be a subscription so the better name would be new-process/[[processName]. Same goes for the other filenames below that contain 'subscription'

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFONewSubscriptionButtonComboBox.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { GET_WORKFLOWS_FOR_DROPDOWN_LIST_GRAPHQL_QUERY } from '../../../graphqlQueries/workflows/workflowsQueryForDropdownList';
+import { PATH_NEW_SUBSCRIPTION } from '../paths';
+import React from 'react';
+import {
+    ComboBoxOption,
+    WFOButtonComboBox,
+} from '../../WFOButtonComboBox/WFOButtonComboBox';
+import { useTranslations } from 'next-intl';
+
+export const WFONewSubscriptionButtonComboBox = () => {
```
DutchBen, line 12: Same comments as above about the naming. I think this should be WFONewProcessButtonComboBox

DutchBen:
Put a todo somewhere in the code for the workflow target filtering? Is there a ticket for that?

ricardovdheijden:
> Put a todo somewhere in the code for the workflow target filtering? Is there a ticket for that?

It is implemented in backend, waiting for deployment. I applied the filtering already in the frontend, right now it results in an empty list until backend is up to date. I put the ticket on blocked

apps/wfo-ui/pages/new-subscription/[processName].tsx
Comment (ricardovdheijden) [reply apps/wfo-ui/pages/new-subscription/[processName].tsx]: Renamed to "start-workflow/[workflowName]"

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFONewSubscriptionButtonComboBox.tsx
ricardovdheijden: Renamed to WFOStartCreateWorkflowButtonComboBox

DutchBen (approved):

PR #301, opened 2023-09-18
WFO Table style update

ricardovdheijden:
Some adjustments to make the table style match better with the design
- gray background table header
- slightly darker hover effect for the rows
- center vertically content in cells

image

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WFOTable/WFOBasicTable/WFOBasicTable.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOTable/WFOBasicTable/WFOBasicTable.tsx
@@ -8,6 +8,8 @@ import {
     WFOTableControlColumnConfig,
     WFOTableDataColumnConfig,
 } from '../utils/columns';
+import { useOrchestratorTheme } from '../../../hooks';
+import { getStyles } from './styles';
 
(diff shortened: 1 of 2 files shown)


packages/orchestrator-ui-components/src/components/WFOTable/WFOBasicTable/WFOBasicTable.tsx
```diff
@@ -8,6 +8,8 @@ import {
     WFOTableControlColumnConfig,
     WFOTableDataColumnConfig,
 } from '../utils/columns';
+import { useOrchestratorTheme } from '../../../hooks';
```
dmarjenburgh, line 11: More of a generic thing, but I think it's a good idea specify imports relative to a base path, or provide path aliases (should be configurable). so you can write imports as

```
import { MyThing } from '@components/things
import { useOrchestratorTheme } from '@hooks'
```

ricardovdheijden: Good one, placed it on our backlog: https://github.com/workfloworchestrator/orchestrator-ui/issues/302

dmarjenburgh (approved):

Issue #232, opened 2023-08-16
Implement time conversion to local machine time and date or time detail


wouter1975:
Backend sends timestamps in ISO format (2023-07-25T08:43:57.383925+00:00). Frontend should show the time/date in local timezone (system time / browser).

Apply this in the whole app 

* On detail page add the timezone.

* On listpages (in tables) just show the time without timezone.
  - [ ] Show only the time when "today"
  - [ ] Show only the date when "not today"

- Determine string format for detail view
- Determine string format for clipboard value

ricardovdheijden:
![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/20791917/f910109b-9d49-40f7-a8d7-4a133ccf9a39)

Using the browsers language settings it is possible to determine the locale of the user

PR #303, opened 2023-09-19
232 date time rendering

ricardovdheijden:
#232 

Improved rendering of dates in the whole app.
- In tables it only shows the time if the event has date "today" and only the date if not today
- In detail views always the full date is shown
- The style of the date 19/09/2023 vs 9-19-2023 etc is derived from the browsers locale settings

image
image
image


packages/orchestrator-ui-components/src/utils/date.ts
```diff
@@ -35,23 +34,44 @@ export const formatDate = (dateString: string | undefined) => {
...
+    }
+
+    return shortNotation
+        ? parseDateToLocaleDateString(date)
+        : parseDateToLocaleDateTimeString(date);
 };
+
+export const parseDateStringRelativeToToday = (
+    dateString: string | undefined,
+    eitherDateOrTime: boolean = false,
```
acidjunk, line 76: this looks funny, `dateOnly`?
It's difficult to read with the boolean here.

It might also be good to incorporate some of the ISO naming in the functions. As a lot of them map into a ISO string representation of a date. And you don't see that when reading the functions.

packages/orchestrator-ui-components/src/utils/date.spec.ts
```diff
@@ -1,11 +1,17 @@
...
+    parseDateToLocaleDateTimeString,
+    parseDateStringRelativeToToday,
     calculateTimeDifference,
+    getCurrentBrowserLocale,
+    parseDateRelativeToToday,
+    parseIsoString,
 } from './date';
 
 const kingsDay2023TimeStamp = 1682589600000;
+const kingsDay = new Date(kingsDay2023TimeStamp);
```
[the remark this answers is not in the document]
acidjunk: 👑

packages/orchestrator-ui-components/src/pages/metadata/WFOProductBlocksPage.tsx
```diff
@@ -137,12 +138,16 @@ export const WFOProductBlocksPage = () => {
         createdAt: {
             field: PRODUCT_BLOCK_FIELD_CREATED_AT,
             name: t('createdAt'),
-            render: parseDateToLocaleString,
+            render: (date) => <WFODateTime dateOrIsoString={date} />,
+            renderDetails: parseIsoString(parseDateToLocaleDateTimeString),
+            clipboardText: parseIsoString(parseDateToLocaleDateTimeString),
         },
         endDate: {
             field: PRODUCT_BLOCK_FIELD_END_DATE,
```
mrijk, line 146: Bit of code duplication in the handling of createdAt/endDate (in multiple locations)

packages/orchestrator-ui-components/src/utils/date.spec.ts
```diff
@@ -29,85 +35,130 @@ describe('date', () => {
...
         it('returns appropriate string if the timedifference is 1 second', () => {
-            const from = new Date(kingsDay2023TimeStamp).toLocaleString();
-            const to = new Date(kingsDay2023TimeStamp + 1000).toLocaleString();
+            const from = new Date(kingsDay2023TimeStamp).toISOString();
+            const to = new Date(kingsDay2023TimeStamp + 1000).toISOString();
             expect(calculateTimeDifference(from, to)).toEqual('00:00:01');
         });
         it('returns appropriate string if the timedifference is 1 minute', () => {
-            const from = new Date(kingsDay2023TimeStamp).toLocaleString();
+            const from = new Date(kingsDay2023TimeStamp).toISOString();
```
mrijk, line 120: I guess parameterized testing (like in pytest) is not supported?

packages/orchestrator-ui-components/src/utils/date.ts
mrijk: I agree on the boolean. Why not just have two functions (with descriptive names) instead of carrying around a boolean parameter (with a default).

packages/orchestrator-ui-components/src/utils/date.spec.ts
ricardovdheijden: Seems to be possible: https://dev.to/tinesoft/be-smart-write-parameterized-tests-with-jest-3eab

packages/orchestrator-ui-components/src/pages/metadata/WFOProductBlocksPage.tsx
ricardovdheijden: This is the config object for the table, therefore we have to tell the table how to render a specific field in case it is an object , like this Date here

packages/orchestrator-ui-components/src/utils/date.ts
ricardovdheijden: Will split this up in 2 functions, that is indeed better

packages/orchestrator-ui-components/src/utils/date.spec.ts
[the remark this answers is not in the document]
ricardovdheijden: It is basically a date to test our functions with, so I will rename it accordingly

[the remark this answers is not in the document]
ricardovdheijden: done

packages/orchestrator-ui-components/src/utils/date.ts
ricardovdheijden: done

acidjunk (approved):

PR #312, opened 2023-09-27
291 forms integration

DutchBen:
https://github.com/orgs/workfloworchestrator/projects/1/views/7?pane=issue&itemId=38494194

Implemented Startworklfow page. Workflows can be started from 
- Create workflows: New subscription button in left menu
- Modify workflows: Subscription detail page

Please note: Not all form buttons and functionality is fully functional yet and are left for future tickets. IMO this is a good set of working functionality but feel free to disagree


packages/orchestrator-ui-components/src/pages/workflow/WFOStartWorkflowPage.tsx
```diff
@@ -22,17 +38,25 @@ export interface UserInputForm {
...
     const { stepUserInput, hasNext } = form;
+    const { stepHeaderStyle, stepListContentBoldTextStyle } = getStyles(theme);
 
     const submit = useCallback(
         (processInput: object[]) => {
             if (workflowName && productId) {
                 processInput.unshift({ product: productId });
             }
+            if (workflowName && subscriptionId) {
+                processInput.unshift({ subscription_id: subscriptionId });
```
ricardovdheijden, line 58: Might be personal preference, but I would like to suggest to use the spread operator over changing the incoming array itself.
const updatedProcessInput = [{ subscription_id: subscriptionId }, ...processInput]

packages/orchestrator-ui-components/src/messages/en-US.json
```diff
@@ -105,7 +107,9 @@
             "viewOptions": "View options",
             "expandAll": "expand all",
             "collapseAll": "collapse all",
-            "duration": "Looptijd"
+            "duration": "Looptijd",
+            "userInput": "User Input",
+            "inProgress": "In progress - Now"
```
ricardovdheijden, line 112: Is this translation correct?

packages/orchestrator-ui-components/src/components/WFOForms/UserInputForm.tsx
```diff
@@ -42,6 +41,12 @@ import ConfirmationDialogContext from '../../contexts/ConfirmationDialogProvider
 import AutoFields from './AutoFields';
 import { ValidationError } from '../../types/forms';
 import { ConfirmDialogActions } from '../../contexts/ConfirmationDialogProvider';
+import { useOrchestratorTheme } from '../../hooks';
+
+type JSONSchemaFormProperty = JSONSchema6 & {
+    uniforms: any;
```
ricardovdheijden, line 47: Cannot be an 'unknown'?

apps/wfo-ui/pages/start-workflow/[workflowName].tsx
```diff
@@ -5,18 +5,21 @@ import { ParsedUrlQuery } from 'querystring';
...
+    const { workflowName, productId, subscriptionId } =
+        router.query as StartWorkFlowPageQuery;
 
-    if (workflowName && productId) {
+    if (workflowName && (productId || subscriptionId)) {
         return (
             <WFOStartWorkflowPage
                 workflowName={workflowName}
                 productId={productId}
+                subscriptionId={subscriptionId}
```
Comment (ricardovdheijden) [inline apps/wfo-ui/pages/start-workflow/[workflowName].tsx:22]: productId or subscriptionId are basically the first payload that will be sent to the backend. In WFOStartWorkflowPage you receive these props and push them into an array to be sent to backend. I think it would be nice to replace the props productId and subscriptionId (only in the WFOStartWorkflowPage component) and replace it with something more generic to just pass the data as is to backend as first step.

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/pages/workflow/WFOStartWorkflowPage.tsx
DutchBen: This doesn't work because it's inside a if/else block unless you make it a let. Made a function that does it that solves the problem

packages/orchestrator-ui-components/src/messages/en-US.json
DutchBen: It's from the design

packages/orchestrator-ui-components/src/components/WFOForms/UserInputForm.tsx
DutchBen: It gives a compile error when i change it. This is legacy code. I think we should change it in a sepearate ticket

apps/wfo-ui/pages/start-workflow/[workflowName].tsx
Comment (DutchBen) [reply apps/wfo-ui/pages/start-workflow/[workflowName].tsx:22]: I'm not sure what you mean?

Comment (DutchBen) [reply apps/wfo-ui/pages/start-workflow/[workflowName].tsx]: I know what you mean now. Made some changes to the typings

Issue #314, opened 2023-09-28
Add polling to process detail page


wouter1975:
Interval 3 sec

PR #334, opened 2023-10-04
314 process detail data polling

ricardovdheijden:
#314 

- Removes useCoreConnection prop from useQueryWithGraphql
- Adds refetchInterval (disables when undefined)

### DIFF
--- a/packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
+++ b/packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
@@ -22,7 +22,6 @@ export const useQueryWithGraphql = <U, V extends Variables>(
         // @ts-ignore
         await graphQLClient.request(query, queryVars);
 
-    // todo : add prop
     return useQuery([queryKey, ...Object.values(queryVars)], fetchFromGraphql, {
         refetchInterval,
     });
--- a/packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
+++ b/packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
@@ -10,17 +10,9 @@ export const useQueryWithGraphql = <U, V extends Variables>(
     query: TypedDocumentNode<U, V>,
     queryVars: V,
     queryKey: string,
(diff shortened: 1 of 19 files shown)


packages/orchestrator-ui-components/src/pages/processes/WFOProcessDetailPage.tsx
```diff
@@ -17,6 +17,8 @@ import {
 } from './timelineUtils';
 import { Step } from '../../types';
 
+const processDetailRefetchInterval = 3000;
```
tjeerddie, line 20: Handy to make it a env variable?

packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
```diff
@@ -29,5 +22,7 @@ export const useQueryWithGraphql = <U, V extends Variables>(
         // @ts-ignore
         await graphQLClient.request(query, queryVars);
 
-    return useQuery([queryKey, ...Object.values(queryVars)], fetchFromGraphql);
+    return useQuery([queryKey, ...Object.values(queryVars)], fetchFromGraphql, {
+        refetchInterval,
+    });
 };
```
dmarjenburgh, line 28: Since `useQueryWithGraphql` is just a small wrapper around `useQuery`, you could consider allowing any `useQuery` options and pass them to `useQuery`. That's would be more flexible than only allowing `refetchInterval` by hand.

dmarjenburgh (dismissed):

packages/orchestrator-ui-components/src/pages/processes/WFOProcessDetailPage.tsx
ricardovdheijden: done

packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
ricardovdheijden: Will implement this when there is a need for the extra props, polling is a temporary solution for now

tjeerddie (approved):

PR #338, opened 2023-10-05
removed loading state

acidjunk:
Seems to be double with work by Ruben


DutchBen:
CHanges are integrated in https://github.com/workfloworchestrator/orchestrator-ui/pull/339

PR #339, opened 2023-10-05
279 subscription general

DutchBen:
- Implemented changes requested in this ticket: https://github.com/workfloworchestrator/orchestrator-ui/issues/279
- Refactored how graphql is called to use the hook that's used in other pages.
- Removed different loading states
- Removed external data loading for now
- Tydied up some types


packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionActions.tsx
```diff
@@ -36,7 +36,7 @@ export type SubscriptionActionsProps = {
     subscriptionId: string;
 };
 
-export const WFOSubscriptionActions: FC<SubscriptionActionsProps> = ({
+export const SubscriptionActions: FC<SubscriptionActionsProps> = ({
```
acidjunk, line 39: Most of the other (sub)components have a WFO prefix; so why is it removed here?

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionDetailTree.tsx
```diff
@@ -0,0 +1,169 @@
+import React, { useState } from 'react';
```
acidjunk, line 1: Why isn't this name WFOSubscriptionDetailTree?

When working with the quick search in my IDE I know have to remember which components have the prefix

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionGeneral.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { useTranslations } from 'next-intl';
+
+import { SubscriptionKeyValueBlock } from './SubscriptionKeyValueBlock';
+import { SubscriptionDetail, KeyValue } from '../../types';
+
+interface SubscriptionGeneralProps {
+    subscriptionDetail: SubscriptionDetail;
+}
+
+export const SubscriptionGeneral = ({
```
acidjunk, line 13: WFO

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionKeyValueBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+                height={20}
+                width={20}
+                color={theme.colors.mediumShade}
+            />
+        );
+
+    return <div>{value as string}</div>;
+};
+
+interface SubscriptionKeyValueBlockProps {
```
acidjunk, line 44: WFO

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionProductBlock.tsx
```diff
@@ -0,0 +1,151 @@
...
+                                                <EuiAccordion
+                                                    id={
+                                                        productBlockInstanceValue.field
+                                                    }
+                                                    arrowDisplay="left"
+                                                    buttonContent={
+                                                        <div>
+                                                            <td
+                                                                valign={'top'}
+                                                                style={{
```
acidjunk, line 123: It might be good to also clean up the inline css

packages/orchestrator-ui-components/src/components/WFOSubscription/WFOSubscription.tsx
```diff
@@ -1,121 +1,44 @@
...
+        `subscription-${subscriptionId}`,
+        true,
+    );
 
-    const onSelectedTabChanged = (id: string) => {
+    const onSelectedTabChanged = (id: SubscriptionTabIds) => {
         setSelectedTabId(id);
     };
 
-    // Todo #97: Find out if pre fetch can be used again. The shape of table cache seems to have changed
```
acidjunk, line 50: This Todo; is still valid.

We had a working cache when navigationg from table to detail page.
Removing it also removes the link to https://github.com/workfloworchestrator/orchestrator-ui/issues/97

packages/orchestrator-ui-components/src/components/WFOSubscription/index.ts
```diff
@@ -1,9 +1,8 @@
 export * from './utils';
-export * from './WFOFixedInputBlock';
 export * from './WFOProcessesTimeline';
-export * from './WFOProductBlock';
-export * from './WFOSubscriptionActions';
-export * from './WFOSubscriptionBlock';
-export * from './WFOSubscriptionDetailTree';
-export * from './WFOSubscriptionGeneral';
+export * from './SubscriptionProductBlock';
```
acidjunk, line 3: Again, I'm fine with renames; but I wonder about the rationale; and it might be good to add it to the ADR/Naming convention rules.

packages/orchestrator-ui-components/src/components/WFOSubscription/utils/utils.tsx
```diff
@@ -1,38 +1,71 @@
...
     {
-        id: RELATED_SUBSCRIPTIONS_TAB,
+        id: SubscriptionTabIds.RELATED_SUBSCRIPTIONS_TAB,
         disabled: true,
         translationKey: 'tabs.relatedSubscriptions',
         prepend: <EuiIcon type="heatmap" />,
     },
 ];
+
+export function getColor(num: number) {
```
acidjunk, line 40: Without loading state : is this till used somewhere?

packages/orchestrator-ui-components/src/messages/en-US.json
```diff
@@ -154,7 +154,21 @@
             "startedBy": "Started by",
             "startedAt": "Started at",
             "status": "Status",
-            "id": "ID"
+            "id": "ID",
```
acidjunk, line 157: Nice; but please keep NL in sync.

And we might want to make this less painful in merges by sorting the keys

acidjunk (changes requested):

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionActions.tsx
DutchBen: The idea is that the WFO prefix is for components that are - or can at some point can be - exposed to the  consumer of the packages and these will not be. Only the complete WFOsubscriptionDeatail component

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionDetailTree.tsx
DutchBen: same

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionGeneral.tsx
DutchBen: same

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionKeyValueBlock.tsx
DutchBen: same

packages/orchestrator-ui-components/src/components/WFOSubscription/index.ts
DutchBen: same comment

packages/orchestrator-ui-components/src/components/WFOSubscription/WFOSubscription.tsx
DutchBen: The issue was a little unclear to me: https://github.com/workfloworchestrator/orchestrator-ui/issues/97 I can put the TODO back but I'm not sure what to do when implementing the ticket

packages/orchestrator-ui-components/src/components/WFOSubscription/utils/utils.tsx
DutchBen: Good point. Will check and remove otherwise

packages/orchestrator-ui-components/src/messages/en-US.json
DutchBen: Updated nl-NL.json

packages/orchestrator-ui-components/src/components/WFOSubscription/utils/utils.tsx
DutchBen: It's not. I removed it

packages/orchestrator-ui-components/src/types/subscription.ts
```diff
@@ -1,43 +0,0 @@
-import { ProductDefinition } from './types';
```
ricardovdheijden, line 1: I think it is better to split the types.ts into different "subjects", just like this file. The types file is getting quite big.

packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -343,3 +325,62 @@ export enum Locale {
...
+    customerId?: string | null;
+    customer?: Customer;
+    externalServices?: ExternalService[];
+};
+
+export type ExternalService = {
+    externalServiceKey: string;
+    externalServiceId: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    externalServiceData: any;
```
ricardovdheijden, line 385: wouldnt "object" be a better choice here?

packages/orchestrator-ui-components/src/components/WFOSubscriptionsList/WFOSubscriptionsList.tsx
```diff
@@ -16,7 +16,7 @@ import {
     WFOTableControlColumnConfig,
     WFOTableWithFilter,
 } from '../WFOTable';
-import { SubscriptionListItem } from './types';
+import { SubscriptionListItem } from '../../types';
```
ricardovdheijden, line 19: I don't think it is a good idea to put all types into the same file. That file is getting large. The reason I place the SubscriptionListItem close to the component is that it is only used here. More specific, a Subscription as we know in the backend should (my opinion) be placed in the types folder, but a so-called view-type that only exists to assist rendering the table should be placed here

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionActions.tsx
ricardovdheijden: I do agree with keeping the WFO prefix since we might want to expose all components and its sub components. A consumer might decide to build his own WFOSubscription with, for example this component.

packages/orchestrator-ui-components/src/types/subscription.ts
DutchBen: I was getting circular dependecies by having these types in a separate file. That's why I put them together

packages/orchestrator-ui-components/src/components/WFOSubscriptionsList/WFOSubscriptionsList.tsx
DutchBen: Placing them in one file is because I was getting circular dependencies as mentioned in another comment. 

On the other part of your argument. I agree with your point on trivial things linke componentProps but a ListItem is general enough to put with the other types imo. If you are looking for types when casually scanning the  package there should be one place to find them

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionKeyValueBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { WFOSubscriptionStatusBadge } from '../WFOBadges';
+import {
+    WFOKeyValueTable,
+    WFOKeyValueTableDataType,
+} from '../WFOKeyValueTable/WFOKeyValueTable';
+import { useOrchestratorTheme } from '../../hooks';
+
+import { KeyValue, SubscriptionStatus } from '../../types';
+
+export const RenderField = (
```
ricardovdheijden, line 15: The RenderField is actually a function returning a component / JSX. Therefore I would suggest to rename this into getRenderField or at least not capitalize the name. (not sure if in scope of PR)

packages/orchestrator-ui-components/src/components/WFOSubscription/WFOSubscription.tsx
DutchBen: Readded the TODO. Can you clarify the ticket a bit?

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionKeyValueBlock.tsx
DutchBen: renamed it to getRenderField

packages/orchestrator-ui-components/src/components/WFOSubscription/SubscriptionProductBlock.tsx
DutchBen: I propose doing this is a separate ticket. There are a bunch of other places where inline css is used so it would be better to do that all at once.

ricardovdheijden (dismissed):

Issue #336, opened 2023-10-05
Open step card when scrolling to card


wouter1975:
When clicking on a dot on the step timeline, the page scrolls to the chosen process step. 

- [ ] add the action to expand the card, and show the state delta.

PR #348, opened 2023-10-06
336 open step card on timeline click

ricardovdheijden:
#336 

- When clicking on timeline item, the step card opens up when scrolling to that step card
- Makes the whole step card header clickable
- Applies round borders to codeblock


packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -50,32 +62,36 @@ export const WFOStepList = React.forwardRef(
...
+                            delta = stateDelta(
+                                previousState,
+                                stepListItem.step.state,
+                            );
                         }
 
-                        // if (step.status === StepStatus.PENDING) {
-                        //     if (step.hasOwnProperty("form")) {
-                        //         Object.keys(step.form.properties as {})
+                        // if (stepListItem.step.status === StepStatus.PENDING) {
```
DutchBen, line 92: Remove this commented code until we need it again

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
```diff
@@ -20,36 +20,53 @@ export const WFOWorkflowStepList = React.forwardRef(
...
+            useState(initialStepListItems);
+
+        const allStepsAreExpanded = stepListItems.every(
+            (item) => item.isExpanded,
         );
 
-        const [stepDetailStates, setStepDetailStates] = useState<
-            Map<number, boolean>
-        >(new Map(allDetailsClosedState));
+        const setExpandedStateStepListItems = (isExpanded: boolean) => {
```
DutchBen, line 34: Nitpicking here but does this function expand all step list items and thus wouldn't expandAllListItems be a more descriptive name?

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
```diff
@@ -20,36 +20,53 @@ export const WFOWorkflowStepList = React.forwardRef(
...
+            );
+        };
 
-        const toggleAllDetailsIsOpen = () => {
-            if (Array.from(stepDetailStates).every((s) => s[1])) {
-                setStepDetailStates(new Map(allDetailsClosedState));
-            } else {
-                setStepDetailStates(new Map(allDetailsOpenedState));
-            }
+        const handleToggleExpandedStepListItem = (
```
DutchBen, line 43: Same here. I would argue that you can toggle and expandedStepListItem, you can collapse it since if you toggle it back the function doesn't do what is called anymore. So just toggleStepListItem?

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
```diff
@@ -58,25 +75,27 @@ export const WFOWorkflowStepList = React.forwardRef(
                     showHiddenKeys={showHiddenKeys}
                     showRaw={showRaw}
                     allDetailToggleText={
-                        Array.from(stepDetailStates).every((s) => s[1])
-                            ? t('collapseAll')
-                            : t('expandAll')
+                        allStepsAreExpanded ? t('collapseAll') : t('expandAll')
```
DutchBen, line 78: I'm not sure if this part of the ticket but we have something about this in the loose ends ticket. I think there should be 2 toggles. One for exand all and another for collapsing. This will show expand all even if there is only one item collapsed although you might still want to collapse all;

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -36,10 +41,17 @@ export const WFOStepList = React.forwardRef(
...
+            scrollToStep: (stepId: string) => {
                 stepReferences.current.get(stepId)?.scrollIntoView({
                     behavior: 'smooth',
-                }),
+                });
+                const foundStepListItem = stepListItems.find(
+                    (value) => value.step.stepId === stepId,
+                );
+                if (foundStepListItem) {
+                    onExpandStepListItem(foundStepListItem);
```
DutchBen, line 52: This functions isn't triggered when the item is expanded but it's actually expading it so a better namen would be expandStepListItem or doExpandStepListItem if you want to keep using verbs.

DutchBen (dismissed):

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
ricardovdheijden: yes, and for now i kept the same logic as it was

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStepList/WFOStepList.tsx
ricardovdheijden: The component "WFOStepList" exposes this function as a prop. Therefore, by convention, we should name it on..... (similar to onClick etc)

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStep/WFOStep.tsx
```diff
@@ -49,24 +47,26 @@ export const WFOStep = React.forwardRef(
...
-            filteredStepDeltaEmpty = true;
-            filteredStepDelta = {};
-            for (const key in delta) {
-                if (!HIDDEN_KEYS.some((word) => key.startsWith(word))) {
-                    filteredStepDeltaEmpty = false;
-                    filteredStepDelta[key] = delta[key];
-                }
-            }
-        }
+        const stepContent = showHiddenKeys
```
mrijk, line 56: Personally I would move this into a separate function and probably get rid of the turnary and replace it by if/else.

Too bad we need this functionality to get rid of some 'random' artifacts like label_ and divider. (btw, what happens if you use just label instead of label_1?) I think we need cleaner data in our components.

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStepList/WFOStepList.tsx
```diff
@@ -50,32 +62,36 @@ export const WFOStepList = React.forwardRef(
...
+                        if (stepListItem.step.status === StepStatus.SUCCESS) {
+                            // Prepare a delta with the last successful stepListItem
                             let previousIndex = index - 1;
                             while (
                                 previousIndex > 0 &&
-                                (steps[previousIndex].status ===
+                                (stepListItems[previousIndex].step.status ===
                                     StepStatus.FAILED ||
-                                    steps[previousIndex].status ===
+                                    stepListItems[previousIndex].step.status ===
```
mrijk, line 79: Maybe assign stepListItems[previousIndex].step.status to a variable? instead of having it twice?

In general I stop reading code as soon as it involves array indices ;) This can almost always be abstracte away with higher-level libs like lodash (or whatever is the flavour of the day) to avoid of-by-one errors.

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
```diff
@@ -20,36 +20,53 @@ export const WFOWorkflowStepList = React.forwardRef(
...
-                            index,
-                            !stepDetailStates.get(index),
-                        ),
-                    ),
-                );
-            }
+        const handleExpandStepListItem = (stepListItem: StepListItem) => {
+            setStepListItems(
+                stepListItems.map((item) => {
+                    if (item.step === stepListItem.step) {
```
mrijk, line 61: This looks quite similar to line 48-53

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
ricardovdheijden: It would SET all isExpanded stepListItems to to given value (true/false). I would call that the expandedState. Have to admit it can be better named. "expandAllListItems" would not cover it all the way since you can also collapse everything with this function. Open for suggestions to improve this name.

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
ricardovdheijden: todo: try to make in line with "setExpandedStateStepListItems"

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStepList/WFOStepList.tsx
ricardovdheijden: rename to triggerExpand....

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStepList/WFOStepList.tsx
ricardovdheijden: Done

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
ricardovdheijden: Done

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStepList/WFOStepList.tsx
ricardovdheijden: renamed to onTrigger...

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStep/WFOStep.tsx
ricardovdheijden: Extracted a function and got rid of the unreadable ternary operator

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOStepList/WFOStepList.tsx
ricardovdheijden: Extracted a previousStep const.

About the array and indices: normally I would agree, but this code is temporary copied from our V1 app. It is waiting to be replaced by a new functionality in the backend to calculate the diff.

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/WFOWorkflowStepList/WFOWorkflowStepList.tsx
ricardovdheijden: extracted a function "updateStepListItem" to get rid of this code duplication

DutchBen (approved):

PR #349, opened 2023-10-06
277 - Task dropdown

acidjunk:
Working start task dropdown + some translations for NL + Fix so menu works on mobile

Implements: https://github.com/workfloworchestrator/orchestrator-ui/issues/277

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WFOButtonComboBox/WFOButtonComboBox.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOButtonComboBox/WFOButtonComboBox.tsx
@@ -7,7 +7,7 @@ import { useOrchestratorTheme } from '../../hooks';
 export type WorkflowComboBoxOption = {
     data: {
         workflowName: string;
-        productId: string;
+        productId?: string;
     };
     label: string;
 };
--- a/packages/orchestrator-ui-components/src/components/WFOButtonComboBox/WFOButtonComboBox.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOButtonComboBox/WFOButtonComboBox.tsx
@@ -1,6 +1,8 @@
 import React, { FC, useState } from 'react';
 import { EuiButton, EuiPopover, EuiSelectable, EuiSpacer } from '@elastic/eui';
 import { getStyles } from './styles';
+import { WFOPlusCircleFill } from '../../icons';
+import { useOrchestratorTheme } from '../../hooks';
 
 export type WorkflowComboBoxOption = {
     data: {
@@ -14,21 +16,31 @@ export type WFOButtonComboBoxProps = {
     buttonText: string;
     options: WorkflowComboBoxOption[];
     onOptionChange: (selectedOption: WorkflowComboBoxOption) => void;
+    isProcess: boolean;
 };
 
 export const WFOButtonComboBox: FC<WFOButtonComboBoxProps> = ({
     buttonText,
     options,
     onOptionChange,
+    isProcess,
 }) => {
     const [isPopoverOpen, setPopoverOpen] = useState(false);
     const { popoverStyle, selectableStyle } = getStyles();
+    const { theme } = useOrchestratorTheme();
 
     const Button = (
         <EuiButton
             onClick={() => setPopoverOpen(!isPopoverOpen)}
-            iconType="plus"
-            fullWidth
+            iconType={
+                isProcess
+                    ? 'plus'
+                    : () => (
+                          <WFOPlusCircleFill color={theme.colors.emptyShade} />
+                      )
+            }
+            fullWidth={isProcess}
+            fill={!isProcess}
         >
             {buttonText}
         </EuiButton>
--- a/packages/orchestrator-ui-components/src/components/WFOButtonComboBox/WFOButtonComboBox.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOButtonComboBox/WFOButtonComboBox.tsx
@@ -26,7 +26,7 @@ export const WFOButtonComboBox: FC<WFOButtonComboBoxProps> = ({
     isProcess,
 }) => {
     const [isPopoverOpen, setPopoverOpen] = useState(false);
-    const { popoverStyle, selectableStyle } = getStyles();
+    const { selectableStyle } = getStyles();
     const { theme } = useOrchestratorTheme();
 
     const Button = (
@@ -48,7 +48,6 @@ export const WFOButtonComboBox: FC<WFOButtonComboBoxProps> = ({
 
     return (
         <EuiPopover
-            css={popoverStyle}
             initialFocus={`.euiSelectable .euiFieldSearch`}
             button={Button}
             isOpen={isPopoverOpen}
--- a/packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React, { FC, useState } from 'react';
 import { EuiSideNav, EuiSpacer } from '@elastic/eui';
 import { useRouter } from 'next/router';
 import {
@@ -17,11 +17,17 @@ import { WFOStartCreateWorkflowButtonComboBox } from './WFOStartCreateWorkflowBu
 
 export const WFOSidebar: FC = () => {
     const router = useRouter();
+    const [isSideNavOpenOnMobile, setisSideNavOpenOnMobile] = useState(false);
+
+    const toggleOpenOnMobile = () => {
+        setisSideNavOpenOnMobile(!isSideNavOpenOnMobile);
+    };
 
     return (
         <EuiSideNav
-            mobileTitle="Nav Items"
-            isOpenOnMobile={false}
+            mobileTitle="Main menu"
+            toggleOpenOnMobile={() => toggleOpenOnMobile()}
+            isOpenOnMobile={isSideNavOpenOnMobile}
             items={[
                 {
                     renderItem: () => (
--- a/packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
+++ b/packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
@@ -1,6 +1,8 @@
-import React, { FC, useState } from 'react';
+import React, { FC, useCallback, useState } from 'react';
 import { EuiSideNav, EuiSpacer } from '@elastic/eui';
 import { useRouter } from 'next/router';
+import { useTranslations } from 'next-intl';
+
 import {
     PATH_METADATA,
     PATH_METADATA_PRODUCT_BLOCKS,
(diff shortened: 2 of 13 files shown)


packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
```diff
@@ -17,11 +17,17 @@ import { WFOStartCreateWorkflowButtonComboBox } from './WFOStartCreateWorkflowBu
 
 export const WFOSidebar: FC = () => {
     const router = useRouter();
+    const [isSideNavOpenOnMobile, setisSideNavOpenOnMobile] = useState(false);
+
+    const toggleOpenOnMobile = () => {
+        setisSideNavOpenOnMobile(!isSideNavOpenOnMobile);
```
DutchBen, line 23: Toggling like this can have unexpected results since the operation to set isSideNavOpenOnMobile might not be finished yet - for example when doing a few in a row.  When setting a value you get the current state so this is my suggestion:

```suggestion
        setisSideNavOpenOnMobile((openState) => !openState);
```

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
```diff
@@ -17,11 +17,17 @@ import { WFOStartCreateWorkflowButtonComboBox } from './WFOStartCreateWorkflowBu
...
+    };
 
     return (
         <EuiSideNav
-            mobileTitle="Nav Items"
-            isOpenOnMobile={false}
+            mobileTitle="Main menu"
```
DutchBen, line 28: Translate?

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
```diff
@@ -0,0 +1,52 @@
...
+            // Avoiding pagination by fetching an unrealistic amount of items
+            first: 1000,
+            after: 0,
+            filterBy: [{ field: 'target', value: 'SYSTEM' }],
+        },
+        'taskWorkflows',
+    );
+
+    const productList: WorkflowComboBoxOption[] = data?.workflows.page.map(
+        (w) => {
```
DutchBen, line 29: Style preference but I prefer longAndDescriptiveNames so workflow instead of w

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
```diff
@@ -0,0 +1,52 @@
...
+            return {
+                label: w.description,
+                data: { workflowName: w.name, productId: '' },
+            };
+        },
+    ) as WorkflowComboBoxOption[];
```
DutchBen, line 35: Are you sure you need this coersion? I would expect it to work without it

packages/orchestrator-ui-components/src/pages/workflow/WFOStartWorkflowPage.tsx
```diff
@@ -35,7 +35,7 @@ type StartWorkFlowPayload =
 
 interface WFOStartWorkflowPageProps {
     workflowName: string;
-    startWorkflowPayload: StartWorkFlowPayload;
+    startWorkflowPayload?: StartWorkFlowPayload | undefined;
```
ricardovdheijden, line 38: ? and undefined is double the same, in case of component props i would go for "?"

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
ricardovdheijden: +1

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
ricardovdheijden: We should actually apply this way in our whole project indeed. Also, officially the app is not supported on mobile. Do we really want to add extra logic specific for this?

packages/orchestrator-ui-components/src/components/WFOButtonComboBox/WFOButtonComboBox.tsx
```diff
@@ -14,29 +16,38 @@ export type WFOButtonComboBoxProps = {
     buttonText: string;
     options: WorkflowComboBoxOption[];
     onOptionChange: (selectedOption: WorkflowComboBoxOption) => void;
+    isProcess: boolean;
```
ricardovdheijden, line 19: If you change this into `isProcess?: boolean` and set the default value to false, you can omit the prop when it is not a process.

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
ricardovdheijden: +1

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
```diff
@@ -0,0 +1,52 @@
...
+        {
+            // Avoiding pagination by fetching an unrealistic amount of items
+            first: 1000,
+            after: 0,
+            filterBy: [{ field: 'target', value: 'SYSTEM' }],
+        },
+        'taskWorkflows',
+    );
+
+    const productList: WorkflowComboBoxOption[] = data?.workflows.page.map(
```
ricardovdheijden, line 28: You also might want to sort the data before showing in the ComboBox

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
acidjunk: I am ok with long names; this was copied form your own code BTW ;)

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
acidjunk: Yes 100% sure

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
acidjunk: No callback? (if you are thinking about better ways; that should be the way forward, I suppose)

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
ricardovdheijden: Placed my comment at the wrong place, here in the correct context:
since `data` is possibly undefined, productList is also possibly undefined. A cast to WorkflowComboBoxOption[] is not recommended. Try to handle undefined or apply a fallback value as empty array: `?? []`

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
DutchBen: What do you mean by no callback?

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
acidjunk: E.g.:

```
    const toggleMobile = useCallback(() => {
        setIsSideNavOpenOnMobile((openState) => !openState);
    }, [setIsSideNavOpenOnMobile])
```
And

```
        <EuiSideNav
            mobileTitle="Main menu"
            toggleOpenOnMobile={toggleMobile}
            isOpenOnMobile={isSideNavOpenOnMobile}
            item=[]
            />          
```

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
acidjunk: Changed it: but now it looks even more distracting (the menu itself is all EN)

packages/orchestrator-ui-components/src/components/WFOButtonComboBox/WFOButtonComboBox.tsx
acidjunk: I know -> I made it this explicit by design.

There will be more places that do stuff for Tasks Or Processes; this way you can always know which one.

I am open for a better name though. We now mix, Task, Process, Workflows and Products in unreliable/inconsistent ways: Consider the `WFOStartCreateWorkflowButton` vs `WFOStartTaskButton` -> They both start a Process...

packages/orchestrator-ui-components/src/pages/workflow/WFOStartWorkflowPage.tsx
acidjunk: Fixed; again silly names here -> StartWorkFlowPayload:

- it starts a process
- All other place use Workflow instead of WorkFlow
- StartWorkFlowPayload can also be used for Modify/Terminate?

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
acidjunk: Fixed and sorted

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
acidjunk: I saw the sort in `WFOStartCreateWorkflowButtonComboBox`:I tried to delegate this to the backend, which seems to be the best choice. Not yet possible to sort on a subresult (products.name)....

So fixed it in JS (it feels a bit slow; you can already open the option box, whilst page is loading/sorting)

packages/orchestrator-ui-components/src/components/WFOStartTaskButtonComboBox/WFOStartTaskButtonComboBox.tsx
```diff
@@ -0,0 +1,53 @@
...
+        'taskWorkflows',
+    );
+
+    const taskList: WorkflowComboBoxOption[] =
+        data?.workflows.page
+            .map((workflow) => {
+                return {
+                    label: workflow.description,
+                    data: { workflowName: workflow.name },
+                } as WorkflowComboBoxOption;
```
ricardovdheijden, line 34: The cast is not needed anymore and you can remove the braces from this oneliner-return statement

packages/orchestrator-ui-components/src/components/WFOPageTemplate/WFOSidebar/WFOSidebar.tsx
DutchBen: My comment was about perusing the possibility of using a function inside of the setState function that gets the current state passed into it instead of using the state value directly to avoid weird errors that result from the order in which React does state updates which are not always sequential. Using or not using arrow function is a valid topic, in this case it doesn't really matter though. The point of doing a state update is to trigger a rerender.

acidjunk: Yes; I agreee -> 2 different situations.

removed the callback construction: as it was just to illustrate my point of view.

1. We have a lot of these arrow functions inside JSX props in our code base
2. We have a lot of function calls who rely on sequential state handling

I left a couple of comments open; in the hope to come up with some rules; it feels like a waste of time that we have to re-discuss these things every PR.

DutchBen (approved):

Issue #318, opened 2023-09-29
Process detail page loose ends


DutchBen:
This tickets is to fix the last missing features of the process detail page. They are
- [x] timelinesteps, only count failures
- [x] scroll into views does not work perfect -- add a wait.
- [x] Flip meaning of Hidden Keys: ON shoould show the keys and OFF should hide them
- [x] Remove full screen from code-block in steps
- [x] Polling: after completing a form and being on the processDetail page the stepCards do not update (timeline is)
- [x] for the first step in the process use the play icon instead of the settings icon, as proposed in design
- [ ] ~Pending step: live time count based on time stamp previous step~
- [x] Add `View options` button TODO: Clarify
  - [x] expanded to choose `Raw JSON data` and `Hidden keys`
    - [x] implement `Raw JSON data` by hiding the step card component, and just stitch all steps states as one json blob.
      - [x] use the same grey design of the step state with the expand and copy functions
      - [ ] ~like Github, new tab with raw JSON~
      - [ ] ~Different from design: raw JSON is not a toggle but a button on the same bar.~
    - [x] implement `Hidden keys` for the card component view, similar as in current GUI.

Skipped:
- [ ] ~Create a 'Expand all' and a 'Collapse all' link. The 'Expand all' link is disabled when all details are expanded, the 'Collapse all' is disabled when all details are collapsed (Original requirement: Implement a toggle function, thus after clicking Expand all, the link is changed to Collapse all with the action to collapse all step cards)~
- [ ] ~introduce polling. https://dev.to/dailydevtips1/automatically-refetching-with-react-query-1l0f~ -- https://github.com/workfloworchestrator/orchestrator-ui/issues/314
- [ ] ~Delta view for modify workflows. << Perhaps an other ticket~

https://projects.invisionapp.com/share/MW134V4MHPB7?hideLogo=true#/screens/471562378

![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/0eb4bb03-7765-4bf3-9276-abafbee54c40)

ricardovdheijden:
Updated the scope of this story. Some items got removed (strike trough)

ricardovdheijden:
![image](https://github.com/workfloworchestrator/orchestrator-ui/assets/20791917/1971c14f-f3be-4a52-b357-fef6e2d7195b)

When a form redirects the user to a processDetail page, the polling mechanism keeps fetching the latest data. All components are getting updated except the steplist

[...]

PR #358, opened 2023-10-12
318 process detail loose ends

ricardovdheijden:
#318 

- [x] timelinesteps, only count failures
- [x] scroll into views does not work perfect -- add a wait.
- [x] Flip meaning of Hidden Keys: ON shoould show the keys and OFF should hide them
- [x] Remove full screen from code-block in steps
- [x] Polling: after completing a form and being on the processDetail page the stepCards do not update (timeline is)
- [x] for the first step in the process use the play icon instead of the settings icon, as proposed in design


packages/orchestrator-ui-components/src/hooks/useWithOrchestratorTheme.ts
```diff
@@ -0,0 +1,9 @@
+import { EuiThemeComputed } from '@elastic/eui';
+import { useOrchestratorTheme } from './useOrchestratorTheme';
+
+export function useWithOrchestratorTheme<T>(
```
DutchBen, line 4: What does the generic type do here? I don't see it being passed in when calling the hook I think

ricardovdheijden: The return type of the getStyles function IS the return type of this function, therefore a generic is used. Calling this hook without specifying the type is acceptable, since the type of getStyles is inferred.

packages/orchestrator-ui-components/src/components/WFOWorkflowSteps/stepListUtils.spec.ts
```diff
@@ -2,25 +2,20 @@ import { getStepContent } from './stepListUtils';
...
+
     it('returns the same object as the input if showHiddenKeys is false', () => {
-        const stepState: StepState = {
-            testProp1: 'testValue1',
-            testProp2: 'testValue2',
-            label_1: 'testLabel1',
-            confirmation_mail: 'testConfirmationMail',
-        };
-        const result = getStepContent(stepState, false);
+        const result = getStepContent(testStepState, true);
```
tjeerddie, line 13: tests seem flipped, this should have false and the one below true

DutchBen (approved):

PR #368, opened 2023-10-16
Status badge tweaks

wouter1975:
- status badges text in lowercase
- productblockbadge now with different colours
- some translation updates for metadata pages
- hidden column for resource type table


packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -63,6 +63,15 @@ export enum ProductLifecycleStatus {
     END_OF_LIFE = 'end of life',
 }
 
+export enum BadgeType {
```
ricardovdheijden, line 66: Some of these are multiple words. Therefore use _ in the names: like PRODUCT_TAG

ricardovdheijden (approved):

Issue #81, opened 2023-06-01
Add filter menu modal per table column


wouter1975:
The initial filter function is based on manual input, which requires understanding of the filter mechanism as described in #76. To enhance the user experience, we would like to add the option to open a modal per column when clicking on the table header cell of the column. This modal offers the option to filter ascending and descending, and it populates a multiselect dropdown to select a filter value (either auto completed string or a specific (sub) string).

From the design the modal (or context menu) should look like
<img width="333" alt="image" src="https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/a6f8e55f-3d15-40e3-8da1-8e70a3da8f3c">

## Limiting scope
In this ticket we are not implementing the dropdown population, just the search input box and the sorting icons

### Implement sorting
From the backend the sortable column information is used to display the sorting icons in the modal. If available implement the correct sorting action (ASC & DESC).

### Implement filtering
Build GUI component with the user actions as described below.

The search input box is the filter input, which will be added to the search bar. If there is already a search string for this particular column in the search bar, the new search string is added (like an OR, thus supporting multi select).

Opening the menu results in the following actions.

- [x] If the column is already sorted, the sorting icon is highlighted
- [ ] ~The unique items list is filled with all unique values from the graphQL endpoint of the table~
  - [ ] ~If the table is already partially filtered, this filter is also applied to the list of unique items~
- [x] the search inputbox is shown, but empty

Result of user actions:
- [x] clicking on sorting icon, will close the menu and sort the table as requested on this column in the chosen direction
- [x] pressing escape or clicking outside the menu, will close the menu, no action is performed on the table
- [ ] typing in the input box ~will further filter the list of unique items.~
  - [ ] if followed by a `Return` ~(and none of the filtered items is selected)~ the menu is closed, the search string is added to the top search bar of the table as described in #76 and as eventhandler of this search bar, the table is updated. (like `customer:mbo`)
  - [ ] ~If clicking on an item in the filtered list the menu, the menu is closed, and the selected filtered item is added to the top search bar of the table as described in #76 and as eventhandler of this search bar, the table is updated. (like `productTag:LP`)~
  - [ ] ~Multiselect: If clicking on an item in the filtered list with the `shift` key, the item is selected.~
  - [ ] ~Multiselect: if one or multiple items have been selected, when pressing `Return` shall update the top search bar with each selected item using the OR operator. (like `productTag:LP OR productTag:IPBGP`)~

[...]

PR #370, opened 2023-10-19
81 table head context menu with free text search

ricardovdheijden:
#81 

This PR does not contain the full scope of the referred story, another PR will.
Scope of this PR:
- Adds popover for each table header
- Adds sort buttons in the popover header, including a small refactor to make direct specific sorting possible)
- Adds text search field and AND's the value to the searchbar above the table (final implementation of this functionality will be in different PR)
- Adds some default sizes to the default-orchestrator-theme. These are copied from the current default theme of EUI. It helps to get some more insight in the theme values.

image
image


packages/orchestrator-ui-components/src/components/WFOTable/WFOBasicTable/WFOTableHeaderCell.tsx
```diff
@@ -1,32 +1,112 @@
...
+    });
+
+    const [isPopoverOpen, setPopover] = useState(false);
+    const handleButtonClick = () => setPopover(!isPopoverOpen);
+    const closePopover = () => setPopover(false);
+
+    const handleChangeSortOrder = (updatedSortOrder: SortOrder) => {
+        if (onSetSortOrder) {
+            onSetSortOrder(updatedSortOrder);
+            closePopover();
```
DutchBen, line 52: Does it make sense to move the closePopover outside of this if? Shouldn't it always close if this function, also on the off chance onSetSortOrder isn't set?

packages/orchestrator-ui-components/src/components/WFOTable/WFOBasicTable/WFOTableHeaderCell.tsx
```diff
@@ -1,32 +1,112 @@
...
+        }
+    };
+
+    const handleSearch = (searchText: string) => {
+        if (onSearch) {
+            onSearch(searchText);
+            closePopover();
```
DutchBen, line 59: Same as above

packages/orchestrator-ui-components/src/components/WFOTable/WFOBasicTable/WFOTableHeaderCell.tsx
```diff
@@ -1,32 +1,112 @@
...
+                onChangeSortOrder={handleChangeSortOrder}
+            />
+        </div>
+    );
+
+    const PopoverContent = () => (
+        <div css={headerCellPopoverContentStyle}>
+            <EuiFieldSearch
+                className={fieldName}
+                placeholder="Search"
```
DutchBen, line 90: Translate?

packages/orchestrator-ui-components/src/icons/WFOSortAsc.tsx
DutchBen, line 1: Add stories so these icons show up in storybook and we maintain a complete set there

packages/orchestrator-ui-components/src/icons/WFOSortDesc.tsx
DutchBen, line 1: Add stories so these icons show up in storybook and we maintain a complete set there

DutchBen:
So the result of this ticket is that you can't sort a column by clicking the arrow anymore and you have to open the modal and change the sorting there. Are we sure that's what we want? I think the extra click doesn't make it better.

packages/orchestrator-ui-components/src/components/WFOTable/WFOBasicTable/WFOTableHeaderCell.tsx
ricardovdheijden: I would say no, because you would give the user the impression that something is happening/changed.

packages/orchestrator-ui-components/src/icons/WFOSortAsc.tsx
ricardovdheijden: Done

packages/orchestrator-ui-components/src/icons/WFOSortDesc.tsx
ricardovdheijden: Done

DutchBen (approved):

ricardovdheijden:
> So the result of this ticket is that you can't sort a column by clicking the arrow anymore and you have to open the modal and change the sorting there. Are we sure that's what we want? I think the extra click doesn't make it better.

Discussed with @wouter1975 , for now there are no plans to keep the single-click-sort. For uniformity we keep the behavior for all tables the same

PR #373, opened 2023-10-19
CIM Service tickets page

Georgi2704:


Georgi2704:
I did an initial setup of the CIM service-ticket page in a seperate surf specific package. Should the structure look like that ? 

I used the subscription components currently just to test it out, will change it soon.

packages/surf/package.json
```diff
@@ -0,0 +1,69 @@
+{
```
ricardovdheijden, line 1: Did you configure this package not to be publishable?

packages/orchestrator-ui-components/src/components/WFOPageTemplate/paths.ts
```diff
@@ -9,3 +9,4 @@ export const PATH_METADATA_RESOURCE_TYPES = '/metadata/resource-types';
 export const PATH_METADATA_WORKFLOWS = '/metadata/workflows';
 export const PATH_TASKS = '/tasks';
 export const PATH_SETTINGS = '/settings';
+export const PATH_SERVICE_TICKETS = '/service-tickets';
```
ricardovdheijden, line 12: After rebasing with main you will get a wfo-ui-surf app next to the standard app. You could add this entry overthere OR in the surf package you created in this PR

apps/wfo-ui/pages/service-tickets/[serviceTicketId].tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
Comment (ricardovdheijden) [inline apps/wfo-ui/pages/service-tickets/[serviceTicketId].tsx:1]: This folder can be moved to surf specific app after rebasing

packages/surf/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "name": "@orchestrator-ui/surf",
```
acidjunk, line 2: Suggest to reflect this naming in the folder.

As it's also done in the orchestrator-ui package.

Or when you don't pulbish to NPM (which we won't do for SURF???) -> we could just use the name "surf"

acidjunk: - Can this new "surf" package use library functions (e.g. stuff from @orchestrator-ui)
- How big will the build of this package be?

@ricardovdheijden / @DutchBen : I don't know if it's a good idea to now have a new "wfo-surf-ui" top level app. I saw it's already merged; but this leads to a lot of duplication and extra maintenance: you now have to sync both folders. At first, till march, this will be only one extra route: "/tickets" which we could also just hide based on an env var.

Personally I would've tried to avoid having 2 duplicate apps as long as possible...

So something to keep in mind when reviewing code. Now commits in "wfo-ui" should be duplicated in "wfo-surf-ui"!

DutchBen: My 2 cents are that because we have this package setup changes that need to be replicated will be very rare or at least rarer than they used to be but good point to keep in mind. I am happy to see that it the setup seems to work.

packages/surf/package.json
Georgi2704: Will adding `"private": true` to package.json be good enough ?

apps/wfo-ui/pages/service-tickets/[serviceTicketId].tsx
Comment (Georgi2704) [reply apps/wfo-ui/pages/service-tickets/[serviceTicketId].tsx]: > Rene:
> @ricardovdheijden / @DutchBen : I don't know if it's a good idea to now have a new "wfo-surf-ui" top level app. I saw it's already merged; but this leads to a lot of duplication and extra maintenance: you now have to sync both folders. At first, till march, this will be only one extra route: "/tickets" which we could also just hide based on an env var.
> 
> Personally I would've tried to avoid having 2 duplicate apps as long as possible...
> 
> So something to keep in mind when reviewing code. Now commits in "wfo-ui" should be duplicated in "wfo-surf-ui"!

Should I wait moving the service-tickets folder to wfo-ui or don't move at all, until a decision is made on the situation above ?

packages/surf/package.json
Georgi2704: Will adding "private": true to package.json be good enough ?

apps/wfo-ui/pages/service-tickets/[serviceTicketId].tsx
Comment (ricardovdheijden) [reply apps/wfo-ui/pages/service-tickets/[serviceTicketId].tsx]: For now we stick to the wfo-ui-surf app next to the normal app to avoid too much SURF specific code in the demo app. You can rebase from main from now on

packages/surf/package.json
ricardovdheijden: Yes that should be good enough

Georgi2704:
Moved everything to #411. Having problems with rebasing and renaming files due to filesystem case sensitivity.

PR #376, opened 2023-10-20
Wfo UI surf app

ricardovdheijden:
Addes a new app for our surf specific pages, with an example page. Had to make changes to the menu bar.
We probably might want to eventually move the collection of providers, currently located in _app.tsx, into the library.

Package Json is adjusted to run either the wfo-ui, wfo-ui-surf or both at the same time

image


Georgi2704:
@ricardovdheijden Should I start my branch for CIM from yours ?

ricardovdheijden:
> @ricardovdheijden Should I start my branch for CIM from yours ?

Maybe after this PR is reviewed. Will try to arrange that this morning.

DutchBen (approved):

PR #377, opened 2023-10-20
Improve collapse and expand of the tree

acidjunk:
The expand of a complete TreeNode and it's children until the next node on the same level works ok.

Todo:
- [x] Fix build
- [x] Add a collapse node
- [x] Clean up some typescript and POC code
- [x] Add some tests


packages/orchestrator-ui-components/src/components/WFOTree/WFOTree.tsx
```diff
@@ -1,16 +1,27 @@
...
-            <WFOTreeBranch key={item.id} item={item} level={0} />
-        ))}
-    </div>
-);
+export const WFOTree: FC<WFOTreeProps> = ({ data, depthList }) => {
+    const { setDepths } = React.useContext(TreeContext) as TreeContextType;
+
+    useEffect(() => {
+        setDepths(depthList);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
acidjunk, line 17: @ricardovdheijden / @DutchBen : I am not satisfied with this eslint ignore -> changes in the TreeContext will cause a re-render; so it's needed here in the deps.

Or do I miss semething? probably the downside of the Frankenstein state management...

packages/orchestrator-ui-components/src/components/WFOTree/styles.ts
```diff
@@ -0,0 +1,16 @@
+export const getStyles = () => {
+    const expandIconContainer = {
+        cursor: 'pointer',
+    };
+
+    const treeContainer = {
+        width: 0,
+        marginTop: 8,
+        marginRight: -8,
```
ricardovdheijden, line 9: Try to use the values from the theme for these.

packages/orchestrator-ui-components/src/components/WFOSubscription/WFOSubscriptionDetailTree.tsx
```diff
@@ -22,36 +22,42 @@ import { WFOSubscriptionProductBlock } from './WFOSubscriptionProductBlock';
 
 type NodeMap = { [key: number]: TreeBlock };
 
-const MAX_EXPAND_ALL = 100;
-
 interface WFOSubscriptionDetailTreeProps {
     productBlockInstances: ProductBlockInstance[];
 }
 
+function getNodeDepth(node: TreeBlock, idToNodeMap: NodeMap): number {
```
ricardovdheijden, line 29: could you add some unit tests for this one?

packages/orchestrator-ui-components/src/components/WFOTree/WFOTree.tsx
ricardovdheijden: You might be able to fix this situation by changing the depthList state (located in the TreeContext) in a useRef.

acidjunk: For now agreed; to try to improve this in upcoming tree tickets: after that decide if we want/need to refactor the Tree context provider.

packages/orchestrator-ui-components/src/components/WFOSubscription/WFOSubscriptionDetailTree.tsx
acidjunk: yes

packages/orchestrator-ui-components/src/components/WfoTree/WfoTreeNode.tsx
```diff
@@ -38,16 +47,19 @@ export const WfoTreeNode: FC<WfoTreeNodeProps> = ({
     }
 
     return (
-        <div style={{ paddingLeft: `${level * 16}px` }}>
+        <div style={{ paddingLeft: `${level * parseInt(theme.size.m)}px` }}>
```
ricardovdheijden, line 50: There is also a base value that you do not need to parse

ricardovdheijden (approved):

DutchBen (approved):

PR #382, opened 2023-10-23
Fix for hot reloading

ricardovdheijden:
Hot reloading was not working anymore. This is a quickfix to start working again. Both apps will always run while developing. Not expecting performance issues on local dev, because the actual compiling kicks in when visiting the page.

### DIFF
--- a/apps/wfo-ui-surf/package.json
+++ b/apps/wfo-ui-surf/package.json
@@ -3,7 +3,7 @@
     "version": "1.0.0",
     "private": true,
     "scripts": {
-        "dev": "next dev",
+        "dev": "next dev -p 3001",
         "test": "jest --passWithNoTests",
         "tsc": "tsc --noEmit",
         "build": "next build",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,9 @@
 {
     "private": true,
     "scripts": {
-        "dev": "turbo run dev",
+        "dev": "turbo run dev --filter=./apps/wfo-ui --filter=./packages/*",
+        "dev:surf": "turbo run dev --filter=./apps/wfo-ui-surf --filter=./packages/*",
+        "dev:all": "turbo run dev",
         "test": "turbo run test",
         "tsc": "turbo run tsc",
         "lint": "turbo run lint",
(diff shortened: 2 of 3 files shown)


apps/wfo-ui-surf/package.json
```diff
@@ -3,7 +3,7 @@
     "version": "1.0.0",
     "private": true,
     "scripts": {
-        "dev": "next dev",
+        "dev": "next dev -p 3001",
```
acidjunk, line 6: This could prove problematic when combined with auth.

The backend allow "*" origin; but I suspect that keycloak will be more strict

package.json
```diff
@@ -1,8 +1,8 @@
 {
     "private": true,
     "scripts": {
-        "dev": "turbo run dev --filter=./apps/wfo-ui",
-        "dev:surf": "turbo run dev --filter=./apps/wfo-ui-surf",
+        "dev": "turbo run dev --filter=./apps/wfo-ui --filter=./packages/*",
```
acidjunk, line 4: is the "./" needed?

I think is also works like this:
--filter=packages/*

acidjunk (dismissed):

package.json
DutchBen: Does this run all packages in packages? Is that what we want?

apps/wfo-ui-surf/package.json
DutchBen: Why fix the port again?

ricardovdheijden: In case you run both apps at the same time with `dev:all` it is a race condition which app runs on port 3000 vs 3001. Fixing the port numbers like this give you more certainty what to expect on each port.

package.json
ricardovdheijden: In this branch/PR yes, when the PR of Georgi gets involved we have to prevent the surf-package will run when we start local dev with the `dev` command

ricardovdheijden: According to one of the examples in the docs, the ./ needs to be there: https://turbo.build/repo/docs/core-concepts/monorepos/filtering#filter-by-directory

DutchBen (approved):

Issue #316, opened 2023-09-28
Table: sort capability per column


ricardovdheijden:
Backend will implement in GraphQL a property in the pagination object that lists the colums that can be sorted on. Once backend is implemented, frontend needs to use this data to make only the sortable columns clickable

PR #384, opened 2023-10-24
316 backend driven sort and filter settings per column

ricardovdheijden:
#316 

Backend response includes information about filterability and sortability per column. This PR uses that information and applies the proper settings to the tableConfig object that os passed into the WFOTable component.

While this PR can already be reviewed, I will only add unit tests for the helper function:
- [x] Unit tests for the helper function

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoProcessesList/WfoProcessList.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoProcessesList/WfoProcessList.tsx
@@ -27,6 +27,7 @@ import {
 } from './processListObjectMappers';
 import { WfoDateTime } from '../WfoDateTime/WfoDateTime';
 import { parseDateToLocaleDateTimeString } from '../../utils';
+import { withSortableValues } from '../WFOTable/utils/withSortableValues';
 
 export type ProcessListItem = Pick<
     Process,
@@ -185,7 +186,7 @@ export const WfoProcessList: FC<WfoProcessListProps> = ({
     if (!data) {
         return <WfoLoading />;
     }
-    const { totalItems } = data.processes.pageInfo;
+    const { totalItems, sortFields } = data.processes.pageInfo;
 
     const pagination: Pagination = {
         pageSize: dataDisplayParams.pageSize,
@@ -201,7 +202,7 @@ export const WfoProcessList: FC<WfoProcessListProps> = ({
     return (
         <WfoTableWithFilter<ProcessListItem>
             data={mapGraphQlProcessListResultToProcessListItems(data)}
-            tableColumns={tableColumns}
+            tableColumns={withSortableValues(tableColumns, sortFields)}
             dataSorting={dataSorting}
             pagination={pagination}
             isLoading={isFetching}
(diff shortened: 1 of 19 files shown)


packages/orchestrator-ui-components/src/components/WFOTable/utils/withSortableAndFilterableValues.ts
```diff
@@ -0,0 +1,23 @@
+import { WFOTableColumns, WFOTableDataColumnConfig } from './columns';
+
+export function withSortableAndFilterableValues<T>(
+    tableColumnConfig: WFOTableColumns<T>,
+    sortableFieldNames: string[],
```
DutchBen, line 5: Can this be keyof T?

packages/orchestrator-ui-components/src/components/WFOTable/utils/withSortableAndFilterableValues.ts
```diff
@@ -0,0 +1,23 @@
+import { WFOTableColumns, WFOTableDataColumnConfig } from './columns';
+
+export function withSortableAndFilterableValues<T>(
```
DutchBen, line 3: I'm not sure about this name. This naming is used by other libraries when they create an Higher Order Component which this one is not. No good alternative come to mind though

packages/orchestrator-ui-components/src/components/WFOTable/utils/withSortableAndFilterableValues.ts
```diff
@@ -0,0 +1,23 @@
...
+    tableColumnConfig: WFOTableColumns<T>,
+    sortableFieldNames: string[],
+    filterableFieldNames: string[],
+): WFOTableColumns<T> {
+    const tableConfigValues =
+        Object.values<WFOTableDataColumnConfig<T, keyof T>>(tableColumnConfig);
+
+    const updatedTableConfigValues = tableConfigValues.map((value) => ({
```
DutchBen, line 11: It feels like this can be l11 and l17 can be done at the same. I'm not sure about how exactly though :)

DutchBen:
Maybe I'm missing something but the sortable and filterable props are now parsed into the table config but I don't see these values used in the actual table implementation to display or not to display anything

packages/orchestrator-ui-components/src/components/WFOTable/utils/withSortableAndFilterableValues.ts
ricardovdheijden: I do want that too, BUT the backend is typing this as string as well. Therefore we (frontend) cannot guarantee what keys will appear here.

packages/orchestrator-ui-components/src/components/WFOTable/utils/withSortableAndFilterableValues.ts
ricardovdheijden: It could, but it might be harder to follow what is happening

packages/orchestrator-ui-components/src/components/WFOTable/utils/withSortableAndFilterableValues.ts
ricardovdheijden: Changing it to mapSortableAndFilterableValuesToTableColumnConfig

packages/orchestrator-ui-components/src/components/WFOTable/utils/withSortableAndFilterableValues.ts
ricardovdheijden: We have a ticket to remove the strict typing : https://github.com/workfloworchestrator/orchestrator-ui/issues/290
Will refer to this discussion, to agree in that ticket what we eventually want to do in the project

DutchBen (approved):

DutchBen (approved):

Issue #238, opened 2023-08-16
Add `Product block` column in `Product blocks`


ricardovdheijden:
Add a column which shows the related (subtending) product blocks, on which this product depend_on. see #248 for backend implementation.

Is implemented, use the `dependsOn` to find the subtending product blocks, and map them in a new column with the name `DependsOn` for shortness. Use badge style.

wouter1975:
created https://github.com/workfloworchestrator/orchestrator-ui/issues/248 --> to discuss whether this should become orchestrator-core issue

wouter1975:
@tjeerddie I think the last update was wrong, I think we are fine with just reporting the dependsOn product blocks. Do you agree?

tjeerddie:
yea, depends on is not a problem because a product block has `ownerSubscriptionId` which shows that the product block is from a different subscription and dependant on the parent. It is not possible yet to get product blocks that use a product block https://github.com/workfloworchestrator/orchestrator-core/issues/383.

example:
```
# productBlockInstances
{
  "id": 0,
  "parent": null,
  "ownerSubscriptionId": "14aa845a-3f11-40b3-8ae0-017343e84eb1"
},
{ # this block depends product block 0
  "id": 1,
  "parent": 0,
  "ownerSubscriptionId": "b159cc09-0719-4895-bfb8-a25d5028fe17"
}
```

PR #389, opened 2023-10-24
Added PB column to metadata PB table

wouter1975:
addresses #238

### DIFF
--- a/packages/orchestrator-ui-components/src/messages/en-US.json
+++ b/packages/orchestrator-ui-components/src/messages/en-US.json
@@ -53,7 +53,7 @@
             "tag": "Tag",
             "status": "Status",
             "resourceTypes": "Resource types",
-            "dependsOn": "Subtending product blocks",
+            "subtendingProductBlocks": "Subtending product blocks",
             "createdAt": "Created",
             "endDate": "End date",
             "parentIds": "Parents"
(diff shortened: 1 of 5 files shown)


packages/orchestrator-ui-components/src/messages/en-US.json
```diff
@@ -53,6 +53,7 @@
             "tag": "Tag",
             "status": "Status",
             "resourceTypes": "Resource types",
+            "dependsOn": "Subtending product blocks",
```
dmarjenburgh, line 56: Never heard of the word `subtending`. Very pretty

ricardovdheijden, line 56: I think it would be nice to have the english translation as close as possible to the key-name. When i openend the page i had to search a bit which column you added to the table. Therefore suggesting to rename `dependsOn` to `subtendingProductBlocks`

packages/orchestrator-ui-components/src/messages/en-US.json
DutchBen, line 1: Add the key to the dutch translation file aswell

wouter1975: check

packages/orchestrator-ui-components/src/messages/en-US.json
wouter1975: check

ricardovdheijden (dismissed):

ricardovdheijden (approved):

PR #410, opened 2023-10-30
401 use wftarget as coloured badge

wouter1975:
- Applies badge style to TARGET in workflow metadata table
- Applies badge style to TARGET in process/task list table

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoBadges/WfoWorkflowTargetBadge/WfoWorkflowTargetBadge.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoBadges/WfoWorkflowTargetBadge/WfoWorkflowTargetBadge.tsx
@@ -11,6 +11,7 @@ export const WfoWorkflowTargetBadge: FC<WfoWorkflowTargetBadgeProps> = ({
     target,
 }) => {
     const { theme, toSecondaryColor } = useOrchestratorTheme();
+    if (!target) return null;
 
     const getBadgeColorFromTarget = (_target: string) => {
         const {
@@ -54,8 +55,6 @@ export const WfoWorkflowTargetBadge: FC<WfoWorkflowTargetBadgeProps> = ({
     };
 
     const { badgeColor, textColor } = getBadgeColorFromTarget(target);
-
-    if (!target) return null;
 
     return (
         <WfoBadge textColor={textColor} color={badgeColor}>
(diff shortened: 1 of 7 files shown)


packages/orchestrator-ui-components/src/components/WfoBadges/WfoWorkflowTargetBadge/WfoWorkflowTargetBadge.tsx
```diff
@@ -0,0 +1,66 @@
...
+                return {
+                    badgeColor: toSecondaryColor(primary),
+                    textColor: primaryText,
+                };
+        }
+    };
+
+    const { badgeColor, textColor } = getBadgeColorFromTarget(target);
+
+    if (target !== null) {
```
DutchBen, line 58: Small style preference. I would to the nullCheck on the target on line 13 so you bail out of the function immediately when there is not target. So 

```if(!target) return null```

**Review by DutchBen — APPROVED** [review]

File: packages/orchestrator-ui-components/src/components/WfoBadges/WfoWorkflowTargetBadge/WfoWorkflowTargetBadge.tsx (hunk 1 above)
Comment (wouter1975) [reply packages/orchestrator-ui-components/src/components/WfoBadges/WfoWorkflowTargetBadge/WfoWorkflowTargetBadge.tsx]: Thanks, I updated code.

File: packages/orchestrator-ui-components/src/components/WfoBadges/WfoWorkflowTargetBadge/WfoWorkflowTargetBadge.tsx [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+                return {
+                    badgeColor: toSecondaryColor(primary),
+                    textColor: primaryText,
+                };
+        }
+    };
+
+    const { badgeColor, textColor } = getBadgeColorFromTarget(target);
+
+    if (!target) return null;
```
DutchBen, line 58: I would move this null check up to line 13 so you bail out right away if there is no target

DutchBen (dismissed):

packages/orchestrator-ui-components/src/components/WfoBadges/WfoWorkflowTargetBadge/WfoWorkflowTargetBadge.tsx
wouter1975: check!

DutchBen (approved):

PR #419, opened 2023-11-02
Honor tree order and started working on hiding keys

acidjunk:
Implementation for https://github.com/workfloworchestrator/orchestrator-ui/issues/282

- [x] Hide empty rows (resource type with no value)
- [x] Move inline styles to styles.ts
- [x] Set default subscription detail tab back to "General"

NOTE: we might want to broaden our linting rules: the last commit contains the results from: `yarn prettier --write .` : which also linted some .jsx stories...

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscription.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscription.tsx
@@ -26,8 +26,7 @@ type WfoSubscriptionProps = {
 export const WfoSubscription = ({ subscriptionId }: WfoSubscriptionProps) => {
     const t = useTranslations('subscriptions.detail');
     const [selectedTabId, setSelectedTabId] = useState<SubscriptionTabIds>(
-        // SubscriptionTabIds.GENERAL_TAB,
-        SubscriptionTabIds.SERVICE_CONFIGURATION_TAB,
+        SubscriptionTabIds.GENERAL_TAB,
     );
 
     const { data, isFetching } = useQueryWithGraphql(
--- a/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscription.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscription.tsx
@@ -26,7 +26,8 @@ type WfoSubscriptionProps = {
 export const WfoSubscription = ({ subscriptionId }: WfoSubscriptionProps) => {
     const t = useTranslations('subscriptions.detail');
     const [selectedTabId, setSelectedTabId] = useState<SubscriptionTabIds>(
-        SubscriptionTabIds.GENERAL_TAB,
+        // SubscriptionTabIds.GENERAL_TAB,
+        SubscriptionTabIds.SERVICE_CONFIGURATION_TAB,
     );
 
     const { data, isFetching } = useQueryWithGraphql(
@@ -52,7 +53,6 @@ export const WfoSubscription = ({ subscriptionId }: WfoSubscriptionProps) => {
             </EuiTab>
         ));
 
-    // Todo #97: Find out if pre fetch can be used again. The shape of table cache seems to have changed
     const subscriptionResult =
         data && data.subscriptions && data.subscriptions.page;
     const subscriptionDetail = subscriptionResult
--- a/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
@@ -82,7 +82,9 @@ export const WfoSubscriptionProductBlock = ({
                                 )}
                             </h3>
                         </EuiText>
-                        <EuiText>Instance ID: {subscriptionInstanceId}</EuiText>
+                        <EuiText>{`${t(
+                            'subscriptionInstanceId',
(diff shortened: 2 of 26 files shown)


packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
```diff
@@ -1,150 +1,174 @@
...
+                                (productBlockInstanceValue) =>
+                                    !HIDDEN_KEYS.includes(
+                                        productBlockInstanceValue.field,
                                     ),
-                            )}
-                        </table>
-                    }
-                </div>
+                            )
+                            .map((productBlockInstanceValue, index) => (
```
mrijk, line 134: Would it make sense to have this as a (function returning a) component instead of deeply nested code?

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscription.tsx
```diff
@@ -26,7 +26,8 @@ type WfoSubscriptionProps = {
 export const WfoSubscription = ({ subscriptionId }: WfoSubscriptionProps) => {
     const t = useTranslations('subscriptions.detail');
     const [selectedTabId, setSelectedTabId] = useState<SubscriptionTabIds>(
-        SubscriptionTabIds.GENERAL_TAB,
+        // SubscriptionTabIds.GENERAL_TAB,
```
DutchBen, line 29: Don't forget to revert back to the GENERAL_TAB when development is done

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
```diff
@@ -1,150 +1,174 @@
...
+                                    style={{
+                                        width: 250,
+                                        paddingLeft: 0,
+                                        paddingTop: 10,
+                                        paddingBottom: 10,
+                                        borderTop: `solid 1px #ddd`,
+                                        borderBottom: `solid 1px #ddd`,
+                                    }}
+                                >
+                                    <b>Owner subscription ID</b>
```
DutchBen, line 111: Translate

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
acidjunk: Done: but translated it to the same :)

packages/orchestrator-ui-components/src/utils/strings.ts
```diff
@@ -8,3 +8,8 @@ export const removeSuffix = (value: string, splitChar = '?'): string => {
     if (value.length <= 1 || !value.includes(splitChar)) return value;
     return value.split(splitChar).slice(0, -1).join();
 };
+
+export const camelToHuman = (value: string): string => {
+    const result = value.replace(/([A-Z])/g, ' $1').trimStart();
```
ricardovdheijden, line 13: Are we certain certain that we don't encounter any numbers? `testTest1234` for example

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
```diff
@@ -1,150 +1,174 @@
...
+                    </EuiFlexItem>
+                    <EuiFlexItem>
+                        <EuiText grow={true}>
+                            <h3>
+                                {getProductBlockTitle(
+                                    productBlockInstanceValues,
+                                )}
+                            </h3>
+                        </EuiText>
+                        <EuiText>Instance ID: {subscriptionInstanceId}</EuiText>
```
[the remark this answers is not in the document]
acidjunk: 🚀

packages/orchestrator-ui-components/src/utils/strings.ts
acidjunk: Numbers should be ok. We might need to check for resource types that start with a number; but I think that's against the python domain model naming (which is the source for the resource types)

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
ricardovdheijden: Currently 204 lines, forgot to push the code?

If yes, stop reading my comment here.

Tbh I prefer even smaller components if possible. There are similar blocks of <td> and <tr> in this component that you could extract into components with a name representing where it stands for.

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
```diff
@@ -1,150 +1,204 @@
...
+                                        valign={'top'}
+                                        css={productBlockLeftColStyle}
+                                    >
+                                        <b>{t('inUseByRelations')}</b>
+                                    </td>
+                                    <td
+                                        valign={'top'}
+                                        css={productBlockRightColStyle}
+                                    >
+                                        <EuiCodeBlock language="json">
```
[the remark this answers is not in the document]
acidjunk: No it isn't because it is styled for process timeline.

Consider this JSON block temporary; to make it easier to debug the backend for inUseByRelations...

PR #424, opened 2023-11-03
422 config with environment variables serverside

ricardovdheijden:
#422 

Using getInitialProps to read the environment variables on runtime. The values of the environment variables will not be compiled into the build anymore.

- [x] Test on dev environment
- [x] Fix one more TODO


DutchBen (approved):

apps/wfo-ui-surf/pages/_app.tsx
```diff
@@ -137,4 +112,12 @@ function CustomApp({ Component, pageProps }: AppProps) {
     );
 }
 
+CustomApp.getInitialProps = async (
```
DutchBen, line 115: I do still wonder if this is not an over complication and if you can't just use this pattern for all values in the initialOrchestratorConig

```    environmentName:
        process.env.NEXT_PUBLIC_ENVIRONMENT_NAME ?? Environment.DEVELOPMENT```

File: packages/orchestrator-ui-components/src/contexts/ApiClientContext.tsx [hunk 2]
```diff
@@ -6,8 +6,9 @@ interface ApiContext {
     apiClient: ApiClient;
 }
 
+// Todo replace this magic string when configuration.ts is moved to package
```
tjeerddie, line 9: don't forget :3

apps/wfo-ui-surf/pages/_app.tsx
ricardovdheijden: According to our test-engineer the current way is higly desired since he does not have to wait long between deploying test environments. We might want to try to make our container image smaller.

packages/orchestrator-ui-components/src/contexts/ApiClientContext.tsx
ricardovdheijden: done

PR #433, opened 2023-11-08
Rollback to hardcoded base URL in the forms api client context

ricardovdheijden:


tjeerddie (approved):

DutchBen (approved):

DutchBen:
I approved but why is this needed? Shouldn't it work?

ricardovdheijden:
> I approved but why is this needed? Shouldn't it work?

It never worked, created a story to fix it: https://github.com/workfloworchestrator/orchestrator-ui/issues/434

PR #463, opened 2023-11-15
434 fix forms api client

ricardovdheijden:


packages/orchestrator-ui-components/src/contexts/ApiClientContext.tsx
```diff
@@ -1,32 +1,25 @@
...
 import { ApiClient, getApiClient } from '../api';
+import { OrchestratorConfigContext } from './OrchestratorConfigContext';
 
 interface ApiContext {
     apiClient: ApiClient;
 }
 
 export const ApiClientContext = createContext<ApiContext>({
-    apiClient: getApiClient('https://orchestrator.dev.automation.surf.net/api'),
+    apiClient: getApiClient('http://localhost:8080/api'),
```
DutchBen, line 11: This value can get set to ''. The default is already set in the config

ricardovdheijden: Done

Georgi2704 (approved):

PR #466, opened 2023-11-17
454 service ticket detail subscription table

Georgi2704:


apps/wfo-ui-surf/components/WfoServiceTicketDetailPage/WfoSubscriptionImpactTable.tsx
```diff
@@ -0,0 +1,226 @@
...
+                },
+            ];
+            itemIdToExpandedRowMapValues[id] = (
+                <EuiDescriptionList listItems={listItems} />
+            );
+        }
+        setItemIdToExpandedRowMap(itemIdToExpandedRowMapValues);
+    };
+
+    const impactTableColumns = {
```
ricardovdheijden, line 88: The type of this const is missing, therefore you might have run into the problems below in WfoBasicTable. The type must be `WfoTableColumns<T>` where T is the type if the data prop in the basic table. When you apply this you can get rid of the ts-expects

apps/wfo-ui-surf/components/WfoServiceTicketDetailPage/WfoSubscriptionImpactTable.tsx
```diff
@@ -0,0 +1,226 @@
...
+            width: '20',
+            isExpander: true,
+            name: (
+                <EuiScreenReaderOnly>
+                    <span>Expand rows</span>
+                </EuiScreenReaderOnly>
+            ),
+            render: (
+                value: keyof ServiceTicketImpactedObject,
+                object: ServiceTicketImpactedObject,
```
ricardovdheijden, line 100: In addition to my comment above: when the type is applied there, you also won't need to specify the type here and in all other render props

apps/wfo-ui-surf/configuration/configuration.ts
```diff
@@ -50,3 +51,15 @@ export const getInitialOrchestratorConfig = (): OrchestratorConfig => {
...
     };
 };
+
+const SURF_CIM_DEFAULT_SENDING_LEVEL = 'resilience_loss';
+
+export const getInitialSurfConfig = (): SurfConfig => {
+    const cimDefaultSendingLevel = process.env.SURF_CIM_DEFAULT_SENDING_LEVEL
+        ? process.env.SURF_CIM_DEFAULT_SENDING_LEVEL
+        : SURF_CIM_DEFAULT_SENDING_LEVEL;
+
```
ricardovdheijden, line 61: You could simplify this with `??` operator

packages/orchestrator-ui-components/src/icons/WfoStatistic.tsx
```diff
@@ -0,0 +1,40 @@
...
+        version="1.1"
+        color={color}
+        xmlns="http://www.w3.org/2000/svg"
+        className="wfoStatisticIcon"
+    >
+        <title>icon/statistic</title>
+        <g xmlns="http://www.w3.org/2000/svg">
+            <path
+                fill={color}
+                fill-rule="nonzero"
```
ricardovdheijden, line 22: Didn't this throw an error? Copying from an SVG file to a React component means you need to replace `fill-rule` with `fillRule` and similar

Georgi2704: will be checked

ricardovdheijden (approved):

PR #470, opened 2023-11-17
Try to get AWS build working

acidjunk:
So this works. Indicating that somehow the new way of providing ENV vars doens't seem to work at all with AWS:

https://test-auth-env.d2zs5rahejck1g.amplifyapp.com/

Schermafbeelding 2023-11-17 om 13 12 02

Todo:
- [ ] find out if it's possible to add runtime env's or find out why they are swallowed.


acidjunk:
@ricardovdheijden found this in the Amplify docs: https://docs.aws.amazon.com/amplify/latest/userguide/ssr-environment-variables.html

So this can probably be fixed by adding some more detailed info to the README. It might be good to describe a couple of deployment strategies. If I understand correctly we can also add an `amplify.yml` file, to the repo. Then it will be supereasy to deploy to AWS again.

PR #484, opened 2023-11-21
ImsPortId and Summary fields

acidjunk:
- [x] Working form field
- [x] Re-add theme support for the Select (try to refactor to SelectField)
- [x] Fix styling
- [x] Try to refactor away the new "Subscription" type
- [x] Also added Summary Field with some styling

Schermafbeelding 2023-11-25 om 05 10 48

Schermafbeelding 2023-11-21 om 22 05 08


packages/orchestrator-ui-components/src/components/WfoForms/formFields/SummaryFieldStyling.ts
```diff
@@ -0,0 +1,83 @@
...
+                borderBottom: `1px solid ${theme.colors.lightestShade}`,
+            },
+        },
+    });
+    return {
+        summaryFieldStyle: summaryFieldStyle,
+    };
+};
+
+// export const summaryFieldStyling = css`
```
ricardovdheijden, line 48: Commented code!

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/api/index.ts
```diff
@@ -160,6 +162,48 @@ export class ApiClient extends ApiClientInterface {
...
+            statusList.join('-'),
+        )}`;
+        const tagsFilter = `tags,${encodeURIComponent(tagList.join('-'))}`;
+        const productsFilter = `products,${encodeURIComponent(
+            productList.join('-'),
+        )}`;
+
+        const params = new URLSearchParams();
+        const filters = [];
+        if (tagList.length) filters.push(tagsFilter);
```
DutchBen, line 191: This is probably legacy but is this intended behaviour? This only pushes to the array if the array is > 0

DutchBen (approved):

packages/orchestrator-ui-components/src/api/index.ts
acidjunk: Function arg:

`tagList: string[] = [],`

Complete logic, copied from V1 indeed:


```
        const statusFilter = `statuses,${encodeURIComponent(
            statusList.join('-'),
        )}`;
        const tagsFilter = `tags,${encodeURIComponent(tagList.join('-'))}`;
        const productsFilter = `products,${encodeURIComponent(
            productList.join('-'),
        )}`;

        const params = new URLSearchParams();
        const filters = [];
        if (tagList.length) filters.push(tagsFilter);
        if (statusList.length) filters.push(statusFilter);
        if (productList.length) filters.push(productsFilter);

        if (filters.length) params.set('filter', filters.join(','));
```

Issue #414, opened 2023-10-31
Sort import by linter


wouter1975:
Would be good to enforce sorted import statements, so files are more easily comparable.

Consider ESlint https://eslint.org/docs/latest/rules/sort-imports or sugestion as proposed in https://dev.to/otamnitram/sorting-your-imports-correctly-in-react-213m

ricardovdheijden:
Opend by Ruben:
https://github.com/workfloworchestrator/orchestrator-ui/pull/472

Will take the config changes to a new branch and then run the fix command again when all branches are closed

ricardovdheijden:
Set an ES-Lint rule according to the config here: https://dev.to/otamnitram/sorting-your-imports-correctly-in-react-213m

Applies rule to all files.

Todos/Considerations

The rule can be applied to files running yarn run lint --fix in the app or packages folders. Currently this doesn't work from the root of the project
There is no integration with prettier. Adding this same rule to prettier requires some more work

acidjunk:
I wonder what the rationale behind the last PR is: eslint is expected to be a "dev" dependency, if I'm not mistaken.

https://withblue.ink/2020/06/07/is-this-a-dependency-or-devdependency.html

PR #498, opened 2023-11-24
414 - Moves all devDependencies into the dependencies section

ricardovdheijden:
#414 

The eslint-config-custom package is used as a devDependency. The dependencies inside the eslint-config-custom should be in the dependencies section to be installed by a consumer of that package


mrijk (approved):

acidjunk:
I wonder what the rationale behind this is. I'm used to have eslint stuff in dev deps.

PR #527, opened 2023-12-01
Subscription Delta

acidjunk:


packages/orchestrator-ui-components/src/components/WfoDiff/WfoDiff.tsx
```diff
@@ -0,0 +1,107 @@
...
+    type: DiffType;
+    hunks: HunkData[];
+}
+
+interface WfoDiffProps {
+    oldText: string;
+    newText: string;
+}
+
+const WfoDiff: React.FC<WfoDiffProps> = ({ oldText, newText }) => {
```
DutchBen, line 33:
```suggestion
const WfoDiff = ({ oldText, newText }: WfoDiffProps) => {
```

packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoWorkflowStepList/WfoWorkflowStepList.tsx
```diff
@@ -17,18 +19,60 @@ export interface WfoWorkflowStepListProps {
     isTask: boolean;
 }
 
+interface SubscriptionDeltaState {
+    subscription: object;
```
DutchBen, line 23: Can we use the Subscription type here?

packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoWorkflowStepList/WfoWorkflowStepList.tsx
```diff
@@ -17,18 +19,60 @@ export interface WfoWorkflowStepListProps {
...
+        data &&
+        'current_state' in data &&
+        'subscription' in (data.current_state as object)
+            ? (data.current_state as SubscriptionDeltaState).subscription
+            : '';
+
+    const newText =
+        data &&
+        'current_state' in data &&
+        '__old_subscriptions__' in (data.current_state as object)
```
ricardovdheijden, line 48: Preferably no casting unless there is really no choice. Similar comment for the 2 ts-ignores below.

DutchBen, line 48: Is it possible to do something like

const currentState = data.current_state as {... __old_state__ .. } so you don't need to use the ts-ignore?

DutchBen: Offcourse but there is a cast here already

packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoWorkflowStepList/WfoWorkflowStepList.tsx
```diff
@@ -17,18 +19,60 @@ export interface WfoWorkflowStepListProps {
...
     const { data, isFetching } = useRawProcessDetails(processId);
     return isFetching ? <WfoLoading /> : <WfoJsonCodeBlock data={data || {}} />;
 };
 
+export const WfoProcessSubscriptionDelta = ({
+    processId,
+}: {
+    processId: string;
+}) => {
+    const { data, isFetching } = useRawProcessDetails(processId);
```
ricardovdheijden, line 36: Maybe a bit out of scope, but still related: why is this hook returning data of type 'object'? I suggest to apply a type inside the useRawProcessDetails hook, at least define a base type that the code in this component relies on. That would resolve the need for type casting below

packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoWorkflowStepList/WfoWorkflowStepList.tsx
acidjunk: I created some type

packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoWorkflowStepList/WfoWorkflowStepList.tsx
acidjunk: Improved it to a solution without casts

acidjunk: Ts ignores are gone.

packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoWorkflowStepList/WfoWorkflowStepList.tsx
acidjunk: It returns a new type now

packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoWorkflowStepList/WfoWorkflowStepList.tsx
```diff
@@ -17,18 +19,50 @@ export interface WfoWorkflowStepListProps {
     isTask: boolean;
 }
 
+// interface SubscriptionDeltaState {
```
ricardovdheijden, line 22: commented code

packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoWorkflowStepList/WfoWorkflowStepList.tsx
```diff
@@ -90,20 +124,24 @@ export const WfoWorkflowStepList = React.forwardRef(
...
                         setExpandedStateStepListItems(!allStepsAreExpanded)
                     }
                     isTask={isTask}
                 />
 
-                {showRaw ? (
-                    <WfoProcessRawData processId={processId} />
-                ) : (
+                {showRaw && <WfoProcessRawData processId={processId} />}
+                {showDelta && (
```
ricardovdheijden, line 141: is there a state possible where Raw and Delta are both shown? Maybe consider introducing an enum with values:
- raw
- delta
- steplist

acidjunk: for now, I am happy with this; if more state changes are needed; I would consider splitting in to separate components.

Issue #543, opened 2023-12-05
Changing search bar string is not reflected in update table


wouter1975:

wouter1975:
If you edit the search bar string manually, and don't confirm with a `Return`, but click outside the inputbox, the search is not executed.

This results in strange behavior, like
* filter on `SPhasjhad`
* table is empty
* remove manually SP from search bar
* click outside the input box
* table remains empty.

ricardovdheijden:
Reserved on my name since i am adusting the current search field

ricardovdheijden:
The searchterm should be submitted when leaving the field (onBlur):
- click outside
- tab out

Issue #478, opened 2023-11-20
Parse search string to BE call and update using filter modals


wouter1975:
This ticket describes the generic behaviour needed to filter tables in a flexible way. Only 1 way sync is required, namely only from filter modal to search bar. In other words, the filter modal has no knowlegde of what is already being filtered for the selected column.

Implement query for tables:
- [x] processes
- [x] subscriptions
- [x] metadata:
   - [x] products
   - [x] product blocks
   - [x] resource types
   - [x] workflows
- [x] tasks

- [ ] Enable the `query` parameter in all graphql queries (currently commented code)
- [x] Implement interactive updating query, using a regex

---

Propose to address this ticket in two separate PRs:
- [ ] first make the search bar to work with the network filter language as described below. The search string is set to the back end, and the table is update accordingly. The current filter modal search string should be simplified to not use " ". 
- [ ] Implement the actual parsing with the logic described below.

---

The search string should comply with the specification as described in:
https://github.com/workfloworchestrator/orchestrator-core/pull/412

Use cases:
* Add search term to existing search string
* Possibility to add search term to an existing column filter
  * effectively addressing multiselect, if the same column_id is already used, add this search term in an `OR` condition, written with the `|` character
* new search term is added to the first column_id match of the existing string
  * even though the search term might be used at multiple places, see example below

Good to know
* `AND` operator is implicit
* `OR` operator is written with ` | `, use spaces around the | character
  * `|` operator is only allowed between `( )` for filtering on multiple value in 1 column
* query is not case sensitive --> just what the user fills in
* just to be 100% clear on the intention, add quotes around string input which contain, `space`, `dash`, `underscore`

see comment from Daniel below

The following table describes the behavior on a subscription detail page:

| Consecutive actions | Resulting in search string |
| -- | -- |
|new search | "" |
| filter column Description on `KLM` | description:KLM |
| filter column Tag on `LP` |  description:KLM tag:LP |  
| filter column Tag on `LR` | description:KLM tag:(LP\|LR) |
| filter column Description on `DESIGN ACA` | description:(KLM\|"DESIGN ACA") tag:(LP\|LR) |
| add search term `1000` in search bar | description:(KLM\|"DESIGN ACA") tag:(LP\|LR) 1000 |
| ~changes in searchtext to OR~ | ~description:(KLM\|"DESIGN ACA")\|tag:(LP\|LR) 1000~ |
[...]

PR #548, opened 2023-12-05
478 filter query frontend update function

ricardovdheijden:
#478 
#543 

- Replace searchbar with a generic one
- Adds query string updater helper
- Renames esQueryString to queryString

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoSearchBar/WfoSearchField.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSearchBar/WfoSearchField.tsx
@@ -1,10 +1,6 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
-import {
-    EuiFormRow,
-    EuiSearchBar,
-    EuiSearchBarOnChangeArgs,
-} from '@elastic/eui';
+import { EuiFieldSearch, EuiFormRow } from '@elastic/eui';
 
 export type WfoSearchFieldProps = {
     __filterQuery?: string; // Deprecated Pythia related way of passing querystring
@@ -22,7 +18,13 @@ export const WfoSearchField = ({
     const queryString = __filterQuery || esQueryString;
     const queryIsValid = true; // Query validation turned of for now until ESQueries can be sent to the backend
 
-    const onChange = ({ queryText }: EuiSearchBarOnChangeArgs) => {
+    const [currentQuery, setCurrentQuery] = useState(queryString ?? '');
+    useEffect(() => {
+        setCurrentQuery(queryString ?? '');
+    }, [queryString]);
+
+    // Todo: clean this up - remove deprecated props
+    const onChange = (queryText: string) => {
         if (__setFilterQuery) {
             __setFilterQuery(queryText);
         } else {
@@ -38,15 +40,16 @@ export const WfoSearchField = ({
             isInvalid={!queryIsValid}
             error={['The query contains invalid parts']}
         >
-            <EuiSearchBar
-                query={queryString}
-                onChange={onChange}
-                box={{
-                    // Todo: possible bug in EUI component EuiSearchBar
-                    // https://github.com/workfloworchestrator/orchestrator-ui/issues/129
-                    // setting the property 'isInvalid' to true has no effect
-                    isInvalid: !queryIsValid,
+            <EuiFieldSearch
+                value={currentQuery}
+                placeholder="Search..."
+                onChange={(event) => setCurrentQuery(event.target.value)}
+                onSearch={(value) => onChange(value)}
+                onBlur={(event) => {
+                    onChange(event.target.value);
                 }}
+                isInvalid={!queryIsValid}
+                fullWidth
             />
         </EuiFormRow>
     );
--- a/packages/orchestrator-ui-components/src/components/WfoSearchBar/WfoSearchField.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSearchBar/WfoSearchField.tsx
@@ -1,5 +1,7 @@
 import React, { useEffect, useState } from 'react';
 
+import { useTranslations } from 'next-intl';
+
 import { EuiFieldSearch, EuiFormRow } from '@elastic/eui';
 
 export type WfoSearchFieldProps = {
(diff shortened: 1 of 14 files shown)


packages/orchestrator-ui-components/src/components/WfoSearchBar/WfoSearchField.tsx
```diff
@@ -1,52 +1,40 @@
...
-                onChange={onChange}
-                box={{
-                    // Todo: possible bug in EUI component EuiSearchBar
-                    // https://github.com/workfloworchestrator/orchestrator-ui/issues/129
-                    // setting the property 'isInvalid' to true has no effect
-                    isInvalid: !queryIsValid,
-                }}
+            <EuiFieldSearch
+                value={currentQuery}
+                placeholder="Search..."
```
DutchBen, line 32: translate

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.ts
```diff
@@ -0,0 +1,52 @@
...
+        return value.slice(1, -1);
+    }
+    return value;
+};
+
+const isContainingSpaces = (value: string) => {
+    return value.indexOf(' ') !== -1;
+};
+
+const toQueryValue = (value: string) => {
```
DutchBen, line 12: IMO quoteSpaces would be more descriptive here because that's all it does.

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.ts
```diff
@@ -0,0 +1,52 @@
+const getValueWithoutBrackets = (value: string) => {
+    if (value.startsWith('(') && value.endsWith(')')) {
+        return value.slice(1, -1);
+    }
+    return value;
+};
+
+const isContainingSpaces = (value: string) => {
```
DutchBen, line 8: Nitpicking but hasSpaces also confirms to the verbNoun pattern of naming

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.ts
```diff
@@ -0,0 +1,52 @@
...
+// Three cases to consider when this function is called
+// 1 - queryString is empty
+// 2 - fieldName already exists in query string
+// 3 - fieldName does not exist in query string
+export const updateQueryString = (
+    queryString: string | undefined,
+    fieldName: string,
+    value: string,
+) => {
+    // 1 - Empty query string returning directly fieldName:value
```
mrijk, line 25: Wondering if these has to be handled as a separate case. Ideally it should be the same as 3? Unless of course queryString can be really undefined (instead of ""), but in that case the comment is inaccurate.

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (!queryString) {
+        return `${fieldName}:${toQueryValue(value)}`;
+    }
+
+    const fieldRegex = new RegExp(
+        `${fieldName}:(\\([\\S "]+\\)|"[\\S ]+"|\\S+)`,
```
mrijk, line 31: - maybe better to use \s instead of space
- You can probably combine the last two as just | [\\S\s]+

dmarjenburgh, line 31: This also includes spaces in the "value" part. It's probably simpler to do the following:

Search for `"${fieldName}:"`
If no match => `//  1`

Otherwise, check if the value is a ValueGroup (starts with `(`)
If it is, find the first closing paren `)` and insert `| toQueryValue(value)` in front of it
If it isn't a ValueGroup, make it a valueGroup by enclosing the existing value in parentheses and insert `| toQueryValue(value)`

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.ts
ricardovdheijden: IMO it does not quote te spaces itself. Therefore i went for this more generic name

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.ts
ricardovdheijden: Done

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        const queryString = '';
+        const fieldName = 'field1';
+        const value = 'value1';
+        expect(updateQueryString(queryString, fieldName, value)).toBe(
+            'field1:value1',
+        );
+    });
+
+    it('adds a new field when the query string is undefined', () => {
+        const queryString = undefined;
```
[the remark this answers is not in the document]
ricardovdheijden: With a plugin that would be possible, but currently not part of our project.

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.ts
ricardovdheijden: Note code is changed a bit, but still a valid question. Skipping the queryString check would mean i get into the `not a match` case and i will get a space in front of the result.

I think it is a bit more readable this (updated) way

packages/orchestrator-ui-components/src/components/WfoTable/WfoTableWithFilter/updateQueryString.ts
ricardovdheijden: Nice one, but are there unforseen characters that we want to catch with `\s`?

Mark90: In python I usually use "raw" strings to avoid having to escape every backslash, for example instead of `"\\d+\\s"` I'd write `r"\d+\s"` (where the 'r' prefix means raw).

It looks like javascript's equivalent for this is `String.raw`. Maybe you can use this to make the part after the ":" slightly more readable?

ricardovdheijden: Updates regex to match backend

dmarjenburgh (approved):

Issue #435, opened 2023-11-08
Implement the `Back` button in user forms


wouter1975:
Check create workflow node.

possibly fixed with #360

PR #556, opened 2023-12-06
435 Adds confirmation dialog when navigating a step back in the form

ricardovdheijden:
#435 #557 

- Adds confirm dialog when jumping back 1 step in forms
- Makes the leave-page-question overridable (without breaking existing implementations)
image

### DIFF
--- a/packages/orchestrator-ui-components/src/messages/en-GB.json
+++ b/packages/orchestrator-ui-components/src/messages/en-GB.json
@@ -48,6 +48,7 @@
             "cancel": "Cancel",
             "submit": "Submit",
             "previous": "Previous",
+            "previousQuestion": "Are you sure to leave this page? Any data filled on this page will be lost",
             "startTask": "Start task",
             "startWorkflow": "Start workflow",
             "resumeTask": "Resume task",
--- a/packages/orchestrator-ui-components/src/messages/en-GB.json
+++ b/packages/orchestrator-ui-components/src/messages/en-GB.json
@@ -48,7 +48,7 @@
             "cancel": "Cancel",
             "submit": "Submit",
             "previous": "Previous",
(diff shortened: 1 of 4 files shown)


packages/orchestrator-ui-components/src/messages/en-GB.json
```diff
@@ -47,6 +47,7 @@
             "cancel": "Cancel",
             "submit": "Submit",
             "previous": "Previous",
+            "previousQuestion": "Are you sure to leave this page? Any data filled on this page will be lost",
```
dmarjenburgh, line 50: I think the below text is a bit cleaner
```suggestion
            "previousQuestion": "Are you sure you want to leave this page? Any filled form data will be lost",
```

packages/orchestrator-ui-components/src/messages/nl-NL.json
```diff
@@ -47,6 +47,7 @@
             "cancel": "Annuleren",
             "submit": "Verstuur",
             "previous": "Vorige",
+            "previousQuestion": "Weet je zeker dat je deze pagina wilt verlaten? Ingevulde velden worden niet opgeslagen",
```
dmarjenburgh, line 50: Nou nou, gaan applicaties nu ook tutoyeren? :stuck_out_tongue: 
```suggestion
            "previousQuestion": "Weet u zeker dat u deze pagina wilt verlaten? Ingevulde velden worden niet opgeslagen",
```

ricardovdheijden: For now that seems to be the way, looking at the other translations. Maybe a good idea to revisit this, since `u` might be better indeed.

DutchBen (approved):

DutchBen (approved):

wouter1975 (approved):

PR #589, opened 2023-12-18
Notification Log tab page

Georgi2704:
Added the page for the Notification Log tab in the service ticket details.

### DIFF
--- a/apps/wfo-ui-surf/components/WfoAvatar/styles.ts
+++ b/apps/wfo-ui-surf/components/WfoAvatar/styles.ts
@@ -1,0 +1,38 @@
+import { EuiThemeComputed } from '@elastic/eui/src/services/theme/types';
+import { css } from '@emotion/react';
+
+export const getStyles = (theme: EuiThemeComputed) => {
+    const stepStateIcon = {
+        height: '42px',
+        width: '42px',
+        alignItems: 'center',
+        display: 'flex',
+        justifyContent: 'center',
+        borderRadius: '24px',
+        border: `2px solid ${theme.colors.emptyShade}`,
+    };
+
+    const openIconStyle = css({
+        ...stepStateIcon,
+        backgroundColor: '#D9EDE1',
+        color: theme.colors.success,
+    });
+
+    const updateIconStyle = css({
+        ...stepStateIcon,
+        backgroundColor: '#CCE3F4',
+        color: theme.colors.primary,
+    });
+
+    const closedIconStyle = css({
+        ...stepStateIcon,
+        backgroundColor: '#DCE4EF',
+        color: theme.colors.subduedText,
+    });
+
+    return {
+        closedIconStyle,
+        updateIconStyle,
+        openIconStyle,
+    };
+};
--- a/apps/wfo-ui-surf/components/WfoAvatar/styles.ts
+++ b/apps/wfo-ui-surf/components/WfoAvatar/styles.ts
@@ -1,32 +1,35 @@
 import { EuiThemeComputed } from '@elastic/eui/src/services/theme/types';
 import { css } from '@emotion/react';
 
-export const getStyles = (theme: EuiThemeComputed) => {
+export const getStyles = (
+    theme: EuiThemeComputed,
+    toSecondaryColor: (color: string) => string,
+) => {
     const stepStateIcon = {
-        height: '42px',
-        width: '42px',
+        height: theme.size.xxl,
+        width: theme.size.xxl,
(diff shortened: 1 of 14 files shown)


packages/orchestrator-ui-components/src/utils/date.ts
```diff
@@ -1,3 +1,5 @@
+import { Locale } from '../types/types';
```
Georgi2704, line 1: A better naming of the types folder or the files inside could be proposed, so they don't clash.

apps/wfo-ui-surf/components/WfoAvatar/styles.ts
```diff
@@ -0,0 +1,38 @@
...
+        alignItems: 'center',
+        display: 'flex',
+        justifyContent: 'center',
+        borderRadius: '24px',
+        border: `2px solid ${theme.colors.emptyShade}`,
+    };
+
+    const openIconStyle = css({
+        ...stepStateIcon,
+        backgroundColor: '#D9EDE1',
```
DutchBen, line 17: Are these colors not part of the theme? This is surf specific but I would still expect everything to follow the general design

packages/orchestrator-ui-components/src/icons/WfoBell.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const WfoBell: FC<WfoIconProps> = ({ width = 20, height = 20 }) => (
+    <svg
+        width={width}
+        height={height}
+        viewBox="0 0 24 24"
+        version="1.1"
+        xmlns="http://www.w3.org/2000/svg"
+    >
+        <title>icon/contact-envelope-fill</title>
```
DutchBen, line 13: This should be icon/bell probably

packages/orchestrator-ui-components/src/utils/date.ts
DutchBen: I have no better suggestions :(. Are you sure you can't use the Locale from the package types?

DutchBen (approved):

packages/orchestrator-ui-components/src/utils/date.ts
Georgi2704: Yep, it fails the unit-tests if use '@/types'

apps/wfo-ui-surf/components/WfoAvatar/styles.ts
Georgi2704: These colors are also used in the WfoWorkflowSteps and they are also not part of the theme.

packages/orchestrator-ui-components/src/icons/WfoBell.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { FC } from 'react';
+
+import { WfoIconProps } from './WfoIconProps';
+
+export const WfoBell: FC<WfoIconProps> = ({ width = 20, height = 20 }) => (
```
ricardovdheijden, line 5: Other icons in this folder are default 24 px, suggesting to do the same here

apps/wfo-ui-surf/translations/en-GB.json
```diff
@@ -68,6 +68,11 @@
                             "extraInformation": "Extra info",
                             "impact": "Impact"
                         }
+                    },
+                    "notificationLog": {
+                        "nl": "NL",
+                        "en": "EN",
+                        "noLogsAvailableForThisServiceTicket": "No logs available for this service ticket"
```
ricardovdheijden, line 75: Did you add tje Dutch translations?

apps/wfo-ui-surf/components/WfoAvatar/styles.ts
```diff
@@ -0,0 +1,38 @@
+import { EuiThemeComputed } from '@elastic/eui/src/services/theme/types';
+import { css } from '@emotion/react';
+
+export const getStyles = (theme: EuiThemeComputed) => {
```
ricardovdheijden, line 4: In addition to the discussion below i would like to suggest to replace all the magic numbers/colors with a value from the theme. Or at least as much as possible

apps/wfo-ui-surf/components/WfoAvatar/styles.ts
wouter1975: would be good to use the theme, maybe also fix this in other parts of the code?

wouter1975 (approved):

PR #594, opened 2023-12-19
577 extra tokens

DutchBen:


Georgi2704 (approved):

packages/orchestrator-ui-components/src/components/WfoSettings/WfoFlushSettings.tsx
```diff
@@ -11,14 +11,19 @@ import {
...
-const clearCache = async (apiUrl: string, settingName: string) => {
+const clearCache = async (
+    apiUrl: string,
+    settingName: string,
+    accessToken: string = '',
+) => {
     const response = await fetch(apiUrl + `/settings/cache/${settingName}`, {
         method: 'DELETE',
+        headers: {
+            Authorization: accessToken ? `Bearer ${accessToken}` : '',
```
tjeerddie, line 25: this could be better done with `useQueryWithFetch` that addes the authorization header, so its not necessary everywhere.

tjeerddie (approved):

packages/orchestrator-ui-components/src/components/WfoSettings/WfoFlushSettings.tsx
DutchBen: Fair enough. I think this should be a separate ticket where we centralise all calls that we do in one place.

PR #603, opened 2023-12-20
579 error handling graphql

DutchBen:
Handles the situation where an error occurs when retrieving data from graphql or during form rendering for example when the eduVPN is inactive

### DIFF
--- a/packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
+++ b/packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
@@ -2,7 +2,10 @@ import { useContext } from 'react';
 import { useQuery } from 'react-query';
 
 import { GraphQLClient } from 'graphql-request';
-import { Variables } from 'graphql-request/build/cjs/types';
+import {
+    GraphQLClientRequestHeaders,
+    Variables,
+} from 'graphql-request/build/cjs/types';
 
 import { TypedDocumentNode } from '@graphql-typed-document-node/core';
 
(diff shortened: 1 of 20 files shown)


packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetailPage.tsx
```diff
@@ -41,6 +42,13 @@ export const WfoProcessDetailPage = ({
         fetchInterval,
     );
 
+    if (error && isError) {
```
acidjunk, line 45: An explanation might make this more readable

packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
```diff
@@ -18,21 +21,26 @@ export const useQueryWithGraphql = <U, V extends Variables>(
...
+        Authorization: session ? `Bearer ${session.accessToken}` : '',
+    };
+
     const fetchFromGraphql = async () =>
         // TS-Ignore because queryVars does not seem to be accepted by the client
         // The props in this useQueryWithGraphql-hook ensures queryVars is indeed related to the query
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
-        await graphQLClient.request(query, queryVars, requestHeaders);
+        await graphQLClient.request<U, V>(query, queryVars, requestHeaders);
```
ricardovdheijden, line 36: You might have fixed the ts-ignore here by adding the types `<U,V>`

ricardovdheijden, line 36: I notice that all pages using this hook now contains a `if error then console log` block. Maybe move that logic into this hook. That keeps the consumers of this hook more clean

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/hooks/useQueryWithGraphql.ts
DutchBen: No. I have tried a lot but wasn't able to remove it.

DutchBen: Done

Issue #99, opened 2023-06-10
Remove backgroundColor in ProductBlock, SubscriptionBlock and FixedInputBlock


acidjunk:
Refactor it so these colors can be provided via the Theme

PR #605, opened 2023-12-20
99 dark mode POC 

acidjunk:
Mostly handy for debugging the places where colors aren't correctly theme-able (badge in header and some other place.)

Also added a color that can be used for the backgrounds of ProductBlocks and Tables + tweaked some cache config.

To make this mergeable:
- [x] refactor local storage part
- [x] decide if we want a provider or use the current prop drilling pattern (for now used prop drilling; probbly good to add the dark mode stuff to the ThemeProvider; so it's easier to use "dark" mode in styles when needed. 
- [x] write some info in the readme about updating EUI (css is now copied from @elatsic/eui/dist folder)


acidjunk:
somewhat related to #99

Issue #578, opened 2023-12-14
Change top page section to dark blue and use new WFO logo


wouter1975:
Logo must decided upon.

- [x] Use the new orchestrator logo
- [x] Use the new color for the top bar (dark blue)
- [x] Use the orchestrator favicon
- [x] Apply the changes to the opensource and surf specific app.

- [x] Ensure that the logo cannot be changed easily
- [x] Text to the right is allowed to be changed

Information of the logo can be found at XXXX

![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/458a5f9f-8b3f-4c54-ab0b-96dd41f3f063)

PR #615, opened 2023-12-28
578 logo update

ricardovdheijden:
#578 

Replaces logo and applies new color to the page header


apps/wfo-ui-surf/components/AppLogo/AppLogo.tsx
```diff
@@ -1,23 +1,27 @@
 import React, { ReactElement } from 'react';
 
-import Image from 'next/image';
+import { EuiText } from '@elastic/eui';
+import { useOrchestratorTheme } from '@orchestrator-ui/orchestrator-ui-components';
 
-import { EuiFlexGroup, EuiFlexItem } from '@elastic/eui';
+import { getAppLogoStyles } from '@/components/AppLogo/styles';
```
DutchBen, line 6: Why are we getting the styles from the library but the logo from the app? I would say both from the same place so either from the library or the app. I would prefer to have the default logo be part of the library

DutchBen (approved):

anieuwland (approved):

PR #636, opened 2024-01-11
536 change tree icon default expanded

wouter1975:
- Change tree icon of root PB
- Use title for root PB in tree instead of PB name
- At page load the root PB is expanded in card view

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoRelatedSubscriptions.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoRelatedSubscriptions.tsx
@@ -86,7 +86,10 @@ export const WfoRelatedSubscriptions = ({
             field: 'description',
             name: t('description'),
             render: (value, record) => (
-                <Link href={`/subscriptions/${record.subscriptionId}`}>
+                <Link
+                    target="_blank"
+                    href={`/subscriptions/${record.subscriptionId}`}
+                >
                     {value}
                 </Link>
             ),
--- a/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
@@ -84,7 +84,7 @@ export const WfoSubscriptionProductBlock = ({
                                 )}
                             </h3>
                         </EuiText>
-                        <EuiText size={'s'}>{`${t(
+                        <EuiText size="s">{`${t(
                             'subscriptionInstanceId',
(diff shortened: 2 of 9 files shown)


packages/orchestrator-ui-components/src/components/WfoTree/WfoTreeNode.tsx
```diff
@@ -44,10 +44,7 @@ export const WfoTreeNode: FC<WfoTreeNodeProps> = ({
     const expanded = expandedIds.includes(item.id);
     const selected = selectedIds.includes(item.id);
 
-    let expandIcon = expanded ? 'arrowDown' : 'arrowRight';
-    if (item.id === 0) {
```
DutchBen, line 48: What did this condition do and are we sure you don't need it anymore? 
It seems it toggles the exapandIcon prop only for the first item which seems significant

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoRelatedSubscriptions.tsx
```diff
@@ -86,7 +86,10 @@ export const WfoRelatedSubscriptions = ({
             field: 'description',
             name: t('description'),
             render: (value, record) => (
-                <Link href={`/subscriptions/${record.subscriptionId}`}>
+                <Link
+                    target="_blank"
+                    href={`/subscriptions/${record.subscriptionId}`}
```
ricardovdheijden, line 91: We hava a constants file with all app paths:
`packages/orchestrator-ui-components/src/components/WfoPageTemplate/paths.ts`

Also I see we did not always use that in the project, just een FYI

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock.tsx
```diff
@@ -84,7 +84,7 @@ export const WfoSubscriptionProductBlock = ({
                                 )}
                             </h3>
                         </EuiText>
-                        <EuiText>{`${t(
+                        <EuiText size={'s'}>{`${t(
```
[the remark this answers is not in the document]
wouter1975: Ok, I will change that throughout the project

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoRelatedSubscriptions.tsx
wouter1975: ok, i will leave it as is.

Issue #544, opened 2023-12-05
Add table export function


wouter1975:
A strong user demand is to have an export function per table.

Usecase for example is that one wants to use a list of services & port of one particular customer.

Requirements:

- [x] export the complete table including hidden columns
- [x] use the applied filterBy and query filter as specified by the user in the FE (or by the url with regards to the filterBy)
- [x] collect the data from all pages (thus none pagination export)
- [x] include table headers for better readability
- [x] use download icon next to the pagination, in right bottom corner -- placed it somewhere else
- [x] limit the export to 1000 rows, to ensure no timeout is hit.

To decide: format choice of .csv or .xls or ... .csv is probably the most generic solution.

ricardovdheijden:
All requirements met, with note: the placement of the button is next to the query bar.

PR #656, opened 2024-01-17
544 table export

ricardovdheijden:
#544 

- Export functionality to export the data from data list pages (subscription list page etc)
- Respects the set filters, but always fetches the first 1000 records
- Always exports all columns available on the page
- Places the export helper funtions in a util file
- Adds a lazy data fetch hook for graphql data -- this prevents that these 1000 records are being fetched on page load.
- not implementing the onExportData prop of the WfoTableWithFilter results in not rendering the Export button to make the export functionality optional

image

### DIFF
--- a/packages/orchestrator-ui-components/src/pages/metadata/WfoWorkflowsPage.tsx
+++ b/packages/orchestrator-ui-components/src/pages/metadata/WfoWorkflowsPage.tsx
@@ -212,6 +212,7 @@ export const WfoWorkflowsPage = () => {
                     getWorkflowsForExport,
                     (data) => data.workflows.page,
                     (data) => data.workflows.pageInfo,
+                    Object.keys(tableColumns),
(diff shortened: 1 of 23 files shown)


packages/orchestrator-ui-components/src/pages/metadata/WfoWorkflowsPage.tsx
```diff
@@ -46,6 +57,8 @@ export type WorkflowListItem = Pick<
 
 export const WfoWorkflowsPage = () => {
     const t = useTranslations('metadata.workflows');
+    const tError = useTranslations('errors');
```
DutchBen, line 60: This seems like a to much repitition between files. Can't some of this be abstracted into a useExport hook? Togheter with the parts where the data is being retrieved?

DutchBen (approved):

wouter1975 (approved):
lgtm

PR #662, opened 2024-01-18
654 npm

DutchBen:
Fix some issues in the setup of our project. Switch from Yarn to NPM


packages/tsconfig/base.json
```diff
@@ -17,7 +17,8 @@
         "preserveWatchOutput": true,
         "skipLibCheck": true,
         "strict": true,
-        "resolveJsonModule": true
+        "resolveJsonModule": true,
+        "target": "es2019"
```
ricardovdheijden, line 21: is this what we want or need?

DutchBen: This option is in the wrong, I have now moved it to "lib". We do need it there after TS began complaining about the presence of WeakMaps. I'm not sure why it didn't complain before

dmarjenburgh (approved):
Very nice

PR #664, opened 2024-01-19
added border radius to wfo tree and empty PB card  placeholder

wouter1975:


packages/orchestrator-ui-components/src/components/WfoTree/WfoTreeNode.tsx
```diff
@@ -86,6 +87,7 @@ export const WfoTreeNode: FC<WfoTreeNodeProps> = ({
                             onClick={() => toggleSelectedId(item.id)}
                             label={item.label}
                             isActive={selected}
+                            style={{ borderRadius: 6 }}
```
ricardovdheijden, line 90: Use the `css` prop instead of `style`.
However it seems be used like that in the Tree component already (non blocking for the PR in this case then)

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoTree/WfoTreeNode.tsx
wouter1975: ok, i leave it as is.

PR #686, opened 2024-01-23
Process table - each column width in absolute or %

wouterhhuisman:
Small improvement to make the column width presentation of process table better on smaller windows.
Change some of the default hidden columns

### DIFF
--- a/.changeset/friendly-poets-compete.md
+++ b/.changeset/friendly-poets-compete.md
@@ -1,0 +1,5 @@
+---
+"@orchestrator-ui/orchestrator-ui-components": patch
+---
+
+Process table - each column width in absolute or %


packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
```diff
@@ -100,18 +100,20 @@ export const WfoProcessesList = ({
...
         lastStep: {
             field: 'lastStep',
             name: t('step'),
+            width: '15%',
         },
         lastStatus: {
             field: 'lastStatus',
             name: t('status'),
-            width: '150',
+            width: '100',
```
ricardovdheijden, line 116: just checking: no percentages? Same for processId

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
wouter1975: Yes, absolute is needed for fix width columns

PR #694, opened 2024-01-24
update date presentation and move getDate to utils

wouter1975:


packages/orchestrator-ui-components/src/pages/startPage/WfoStartPage.tsx
```diff
@@ -90,7 +92,9 @@ export const WfoStartPage = () => {
         listItems:
             processesSummaryResult?.processes.page.map((workflow) => ({
                 title: workflow.workflowName,
-                value: workflow.startedAt,
+                value: parseDateToLocaleDateTimeString(
```
ricardovdheijden, line 95: Just checking: have you tried `formatDate(...)` -- I think that one does getDate and parseDate together.
Can be found in `date.ts`

wouter1975: I tried but not good enough, I throught it did not support the string types, but I retried and works fine. I will update the code.

packages/orchestrator-ui-components/src/pages/startPage/WfoStartPage.tsx
```diff
@@ -68,10 +68,10 @@ export const WfoStartPage = () => {
         headerStatus: SummaryCardStatus.Neutral,
         listTitle: t('activeSubscriptions.listTitle'),
         listItems:
-            subscriptionsSummaryResult?.subscriptions.page.map(
+            subscriptionsSummaryResult?.subscriptions?.page.map(
```
ricardovdheijden, line 71: As far as i know is `subscriptions` not "possibly undefined". Same goes for the other cases in this PR.

DutchBen (approved):

packages/orchestrator-ui-components/src/pages/startPage/WfoStartPage.tsx
```diff
@@ -71,7 +71,7 @@ export const WfoStartPage = () => {
             subscriptionsSummaryResult?.subscriptions.page.map(
                 (subscription) => ({
                     title: subscription.description,
-                    value: getFirstUuidPart(subscription.subscriptionId),
+                    value: formatDate(subscription?.startDate),
```
ricardovdheijden, line 74: The possibly undefined comment applies here too. The `subscription`, `workflow` and `task` cannot be undefined

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/pages/startPage/WfoStartPage.tsx
wouter1975: ok

PR #715, opened 2024-01-30
431 rtk fixes

DutchBen:


apps/wfo-ui/next.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-    reactStrictMode: false,
+    reactStrictMode: true,
```
ricardovdheijden, line 2: FYI. We had to disable this in the past because of EUI. Are all components verified that they still work?

packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
```diff
@@ -236,9 +236,13 @@ export const WfoProcessesList = ({
     return (
         <WfoTableWithFilter<ProcessListItem>
             queryString={queryString}
-            data={mapGraphQlProcessListResultToProcessListItems(
-                processes || [],
-            )}
+            data={
+                !isFetching
```
ricardovdheijden, line 240: This should not be needed. Why would it be needed

ricardovdheijden (approved):

Issue #700, opened 2024-01-25
Fix crash and use the new MINL from CIM when expanding row in the impact table


Georgi2704:
![image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/61829525/362398ca-aead-46fe-a30c-4dd54bcc30de)

Use the new customer_minl field inside the impacted customers table and fix the crash that currently happens when you expand a row.

PR #718, opened 2024-01-30
Fix for impact table expanding row and show inform customers

Georgi2704:
fixes https://github.com/workfloworchestrator/orchestrator-ui-library/issues/700

### DIFF
--- a/apps/wfo-ui-surf/types/types.ts
+++ b/apps/wfo-ui-surf/types/types.ts
@@ -163,14 +163,14 @@ export interface ImpactedObject {
     related_customers: ImpactedRelatedCustomer[];
 }
 
-export type ImpactTableColumns = Pick<
-    ImpactedObject,
-    'subscription_id' | 'subscription_description' | 'impact_override'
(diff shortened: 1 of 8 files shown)


DutchBen (approved):

apps/wfo-ui-surf/types/types.ts
```diff
@@ -109,20 +109,20 @@ export interface ServiceTicketLog {
 }
 
 export enum ImpactLevel {
-    NO_IMPACT = 'no_impact',
+    NEVER = 'no_impact',
```
ricardovdheijden, line 112: is this supposed to match?

ricardovdheijden (approved):

apps/wfo-ui-surf/types/types.ts
Georgi2704: For now I'll leave it like this, but I am planning to discuss it with Wouter or someone from the backend team, because there is some mismatch in the backend

PR #734, opened 2024-01-31
Update 'productTag' field to 'tag'

### LINKED ISSUE #547: Add option to clear/reset the TSV index
The TSV column is critical in the filter options of the subscription table. It the TSV is not being updated anymore, or corrupt, it will be very useful for the user to quickly reset this column, like a cache clear.

- [ ] ~~add a option in the clear cache dropdown~~
- [ ] or, add a new action button to trigger this TSV reset.
  - To reset the search-index, make a POST to: api/settings/search-index/reset
  - On success, it will return a status 200 response (no payload).
- [ ] ~~add a new dropdown item in de clear cache settings~~ --> probably not the best place as the values in the dropdown are driven by the backend.
  - ~~Text in dropdown "Clear all text search indexes from subscriptions table"~~

- [ ] ~~add this function to the subscription page with the button text: `Reset search index`~~
  -  ~~at the same place as the Retry buttons on the task page.~~

### LINKED ISSUE #690: [MEDIUM] Change productTag column in process table to tag
To become consistent, change `productTag` into tag.
With this change it is aligned with the `tag` in subscription table.

The result will be that the query string when filter in the tag column will become: `tag:....`

This will require a backend change, see github: https://github.com/workfloworchestrator/orchestrator-core/issues/473

dmarjenburgh:
Resolves #690

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
@@ -59,7 +59,7 @@ export type ProcessListItem = Pick<
     startedAt: Date;
     lastModifiedAt: Date;
     productName?: string;
-    productTag?: string;
+    tag?: string;
     customer: string;
     customerAbbreviation: string;
 };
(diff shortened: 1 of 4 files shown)


DutchBen:
Nice! It's an often used convention to prefix the branch name with the ticket number so 640-update-producttag-to-tag

DutchBen:
I noticed 'productTag' is still used in the workflowList page. Should that be included here to?

DutchBen (approved):

ricardovdheijden (approved):

PR #747, opened 2024-02-05
719: Remove turbo eslint rules so downstream pacakges dont fail them

DutchBen:
Remove turbo as standard extend option for eslint configurations so builds usingt this package wont fail if they dont use turbo

### DIFF
--- a/packages/eslint-config-custom/index.js
+++ b/packages/eslint-config-custom/index.js
@@ -3,7 +3,6 @@ module.exports = {
         'plugin:@typescript-eslint/recommended',
         'next',
         'next/core-web-vitals',
-        'turbo',
         'prettier',
     ],
     rules: {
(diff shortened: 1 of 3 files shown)


packages/eslint-config-custom/index.js
```diff
@@ -3,7 +3,6 @@ module.exports = {
         'plugin:@typescript-eslint/recommended',
         'next',
         'next/core-web-vitals',
-        'turbo',
```
ricardovdheijden, line 6: You might also want to remove the dependency: https://github.com/workfloworchestrator/orchestrator-ui-library/blob/main/packages/eslint-config-custom/package.json#L14

DutchBen: Done

tjeerddie (approved):

ricardovdheijden (approved):

Issue #723, opened 2024-01-31
[HIGH] FW summary page is not shown, fatal error


wouter1975:
In FW last page cannot be loaded, ui fails in a hard way. This error is shown in the console.

![Image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/83226590/8dabdadd-6d85-461e-a491-bef6995a724a)

PR #748, opened 2024-02-05
723: Fix SummaryField .includes is not a function

tjeerddie:
- check if row is string before trying .includes.
- change incorrect default customerKey in ContactPersonName field

Fixes: #723


DutchBen (approved):

tjeerddie:
fixed pushes as incorrect user

ricardovdheijden (approved):

Issue #683, opened 2024-01-23
[HIGH] Port dropdown not prefilled in modify WF


wouter1975:
On the BGP page we have to select one or more ports on the first page. Then we get additionally pages for each port. However the port is not "pre filled" in the dropdown fields on these pages. I cannot select it either because on these "port specific pages" the port filed is configured as read-only.

- [ ] Also applies to the create workflow of IP BGP - the WF can be started however, but for the user is is not clear which port you are configuring
- [x] Most probably fixes also L3VPN, where no ports can be selected

![Image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/83226590/f4e9a59c-0b37-462b-9760-4b2fd8cea332)

PR #750, opened 2024-02-05
683-dropdown-bug: Use value even if dropdown is disabled

DutchBen:


tjeerddie (approved):
LGTM and connect it to the issue: #683

wouter1975 (approved):

ricardovdheijden (approved):

PR #753, opened 2024-02-06
751 Add authToken to queries send through RTK

DutchBen:


packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -26,11 +38,17 @@ export const orchestratorApi = createApi({
             case BaseQueryTypes.fetch:
                 const fetchFn = fetchBaseQuery({
                     baseUrl: orchestratorApiBaseUrl,
+                    prepareHeaders: (headers) => {
```
tjeerddie, line 41: I don't think the function is necessary here and you can just do `prepareHeaders: prepareHeaders` instead.

tjeerddie (approved):
LGTM

packages/orchestrator-ui-components/src/rtk/api.ts
ricardovdheijden: think so too and if key value have the same name `prepareHeaders,` is sufficient

packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -13,6 +17,14 @@ type ExtraOptions = {
     baseQueryType?: BaseQueryTypes;
 };
 
+const prepareHeaders = async (headers: Headers) => {
+    const session = (await getSession()) as SessionWithToken;
+    if (session && session.accessToken) {
```
ricardovdheijden, line 22: this could be session?.accessToken

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/rtk/api.ts
DutchBen: Done

packages/orchestrator-ui-components/src/rtk/api.ts
DutchBen: Done

Issue #685, opened 2024-01-23
[HIGH] In the "replace port" workflow for IP BGP the form keeps reloading


wouter1975:
In the "replace port" workflow for IP BGP the form keeps reloading therefore I am unable to enter any values in the input elements.

Zie bijv: 
https://orchestrator-v2.dev.automation.surf.net/workflows/946c101e-17bb-4f3f-a9e7-9d623d528243

Issue #650, opened 2024-01-16
WF form is refreshed every couple of seconds, loosing information


wouter1975:

tjeerddie:
same problem as #685

PR #754, opened 2024-02-06
685: Fix form reloading and resetting inputs

tjeerddie:
- change process const to state var in WfoProcessDetailPageProps and check with lastStatus or lastStep is different if the process actually needs an update.

Fixes: #685, #650

### DIFF
--- a/packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetailPage.tsx
+++ b/packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetailPage.tsx
@@ -8,7 +8,12 @@ import {
 } from '../../components/WfoWorkflowSteps';
 import { GET_PROCESS_DETAIL_GRAPHQL_QUERY } from '../../graphqlQueries';
 import { useQueryWithGraphql } from '../../hooks';
-import { ProcessDoneStatuses, ProcessStatus, Step } from '../../types';
+import {
+    ProcessDetail,
+    ProcessDoneStatuses,
+    ProcessStatus,
+    Step,
+} from '../../types';
 import { getProductNamesFromProcess } from '../../utils';
 import { WfoProcessDetail } from './WfoProcessDetail';
 import {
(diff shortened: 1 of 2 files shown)


packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetailPage.tsx
```diff
@@ -64,7 +70,22 @@ export const WfoProcessDetailPage = ({
...
+        }
+
+        const updateProcess =
+            process.lastStatus != fetchedProcessDetails?.lastStatus ||
+            process.lastStep !== fetchedProcessDetails?.lastStep;
+        if (updateProcess) {
+            setProcess(fetchedProcessDetails);
+        }
+    }, [data]);
+
```
ricardovdheijden, line 88: I think it might be better to disable the refetching from the useQueryWithGraphql hook instead of introducing more state.

packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetailPage.tsx
```diff
@@ -64,7 +70,22 @@ export const WfoProcessDetailPage = ({
...
+            return;
+        }
+
+        const updateProcess =
+            process.lastStatus != fetchedProcessDetails?.lastStatus ||
+            process.lastStep !== fetchedProcessDetails?.lastStep;
+        if (updateProcess) {
+            setProcess(fetchedProcessDetails);
+        }
+    }, [data]);
```
DutchBen, line 87: Add the process dependency as suggested below

packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetailPage.tsx
DutchBen: This doesn't disable refetching though right? It only determines if an update should take place

tjeerddie: every interval the data updates and is seen as a state change, updating everything. this doesn't disable fetching interval but changes it so the form components don't rerender each interval.

ricardovdheijden (approved):

PR #778, opened 2024-02-13
use GitHub app generated token for repository access

hanstrompert:
Use a GitHub app on the workflow orchestrator organisation with Content and Metadata access to generate tokens that have access to the orchestrator-ui-library and example-orchestrator-ui repositories. This way there is no need to use personal access tokens, and the creation of a tag can trigger another workflow in the same repository.


tjeerddie (approved):
LGTM

DutchBen (approved):

PR #821, opened 2024-02-21
Changes needed for the cim send email form wizard

Georgi2704:


packages/orchestrator-ui-components/src/api/index.ts
```diff
@@ -24,9 +24,9 @@ import { ProductDefinition } from '@/types';
 
 import { getAxiosInstance } from './axios';
 
-const CIM_FORMS_ENDPOINT = 'surf/cim/forms/';
 const PROCESS_ENDPOINT = 'processes/';
 const PRODUCTS_ENDPOINT = 'products/';
+const FORMS_ENDPOINT = 'surf/forms/';
```
DutchBen, line 29: This is Surf specific no? Is there no way to get this in the gitlab version? For example by import the BaseClient and extending it there?

packages/orchestrator-ui-components/src/utils/strings.ts
```diff
@@ -13,3 +13,12 @@ export const camelToHuman = (value: string): string => {
     const result = value.replace(/([A-Z])/g, ' $1').trimStart();
     return result.charAt(0).toUpperCase() + result.slice(1);
 };
+
+export const snakeToHuman = (value: string): string => {
```
DutchBen, line 17: Use ChatGtp to create some unit tests for these functions

DutchBen (approved):

packages/orchestrator-ui-components/src/api/index.ts
ricardovdheijden, line 29: surf specific things should be placed in the surf specific app repository

Georgi2704: This is for the CreateForm endpoint that is also in wfo-ui. I agreed with Ruben to not delete it for now.

packages/orchestrator-ui-components/src/utils/string.spec.ts
```diff
@@ -68,3 +74,33 @@ describe('camelToHuman()', () => {
         expect(result).toEqual('A Quick Brown Fox');
     });
 });
+
+describe('snakeToHuman()', () => {
+    it("Doesn't crash on an empty string but returns empty string", () => {
```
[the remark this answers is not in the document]
Georgi2704: I looked from the examples above

ricardovdheijden (approved):

PR #843, opened 2024-02-28
Added isDisabled prop to WfoSubmitModal

Georgi2704:


packages/orchestrator-ui-components/src/components/WfoSettingsModal/WfoSubmitModal.tsx
```diff
@@ -18,6 +18,7 @@ export type WfoSubmitModalProps = {
     onClose: () => void;
     onSubmit: () => void;
     isLoading: boolean;
+    isDisabled?: boolean;
```
ricardovdheijden, line 21: Looking at the usage of this prop it seems you only prevent the submit part. You can still cancel. Suggesting to rename this prop to `isSubmitDisabled` or something that explains what is being disabled

ricardovdheijden (approved):

PR #875, opened 2024-03-06
Override RTK baseQuery by providing a custom baseQuery

Georgi2704:


packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -34,12 +35,21 @@ const prepareHeaders = async (headers: Headers) => {
...
-        const { baseQueryType } = extraOptions || {};
+        const { baseQueryType, customBaseQueryType } = extraOptions || {};
 
         const state = api.getState() as RootState;
         const { orchestratorApiBaseUrl, graphqlEndpointCore } =
             state.orchestratorConfig;
 
+        const customBaseQueries = state.customBaseQueries;
+
+        if (customBaseQueries) {
```
ricardovdheijden, line 46: This can be simplified, basicly you are performing a `find` here in an array that could also be undefined

apps/wfo-ui-surf
ricardovdheijden, line 1: This might be a mistake, there should be no references of submodules refering to wfo-ui-surf

packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -21,9 +21,10 @@ export enum CacheTags {
 
 type ExtraOptions = {
     baseQueryType?: BaseQueryTypes;
+    customBaseQueryType?: string;
```
ricardovdheijden, line 24: The name suggests that this could be an enum or more stongly typed

packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -34,12 +35,21 @@ const prepareHeaders = async (headers: Headers) => {
...
 
         const state = api.getState() as RootState;
         const { orchestratorApiBaseUrl, graphqlEndpointCore } =
             state.orchestratorConfig;
 
+        const customBaseQueries = state.customBaseQueries;
+
+        if (customBaseQueries) {
+            for (const query of customBaseQueries) {
+                if (query.queryType === customBaseQueryType) {
```
DutchBen, line 48: Maybe we should add a check and an error message for the case when there is a customBaseQuery and a queryType but the queryType doesn't match any of the customBaseQueris and none of the BaseQueryTypes. Currently that means it will catch the default clause and run it as a graphQlQuery with the default url. 

It might be useful to have an error when you misttype or forget to implement something here

packages/orchestrator-ui-components/src/rtk/slices/customBaseQueries.ts
```diff
@@ -0,0 +1,31 @@
...
+    FetchBaseQueryMeta
+>;
+export type CustomBaseQuery = {
+    queryType: string;
+    customFn: CustomBaseQueryFn;
+};
+type CustomBaseQueriesSlice = Slice<CustomBaseQuery[]>;
+
+export const getCustomBaseQueriesSlice = (
+    config: CustomBaseQuery[],
```
DutchBen, line 24: I think you should call this customeBaseQueries instead of config

packages/orchestrator-ui-components/src/rtk/store.ts
```diff
@@ -16,18 +20,26 @@ export type RootState = {
     >;
     toastMessages: ReturnType<typeof toastMessagesReducer>;
     orchestratorConfig: OrchestratorConfig;
+    customBaseQueries?: CustomBaseQuery[];
 };
 
 export const getOrchestratorStore = (
     orchestratorConfig: OrchestratorConfig,
+    customBaseQueries?: CustomBaseQuery[],
```
DutchBen, line 28: I think I prefer for this to have a default value of []. so below on line 40 you can just always add it but having it empty by default

DutchBen (approved):

ricardovdheijden (approved):

Issue #702, opened 2024-01-25
Move all Graphql queries to RTKQuery


DutchBen:
Move all graphql queries that currently still use graphQLClient to use RTKQuery

Todo:
- replace magic value for number of results in [packages/orchestrator-ui-components/src/pages/startPage/WfoStartPage.tsx](https://github.com/workfloworchestrator/orchestrator-ui-library/pull/902/files#diff-3e4ee0aec957a400f858ee7e47fc03a09cd1aaeaec340328b2dc7feee6ef5cee) -- waiting for merging #909 

Dividing the work a bit:
@DutchBen 
- relatedSubscriptionsQuery (https://github.com/workfloworchestrator/orchestrator-ui-library/pull/901)
- workflows/workflowsQuery
- workflows/workflowsQueryForDropdownList

@ricardovdheijden:
- subscriptionsDropdownOptionsQuery (https://github.com/workfloworchestrator/orchestrator-ui-library/pull/909)

Done:
- processDetail (merged)
- ProcessListQuery (merged)
- resourceTypeQuery (merged)
- ProcessStepsQuery (merged)
- subscriptionList(page + summaryCard)
- productBlocksQuery
- subscriptionDetail (PR: https://github.com/workfloworchestrator/orchestrator-ui-library/pull/896)
- productsQuery (https://github.com/workfloworchestrator/orchestrator-ui-library/pull/902)

PR #886, opened 2024-03-08
702 subscription list query

ricardovdheijden:
#702  SubscriptionList query


packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionList.ts
```diff
@@ -0,0 +1,86 @@
...
+            GraphqlQueryVariables<SubscriptionListItem>
+        >({
+            query: (variables) => ({
+                document: subscriptionListQuery,
+                variables,
+            }),
+            transformResponse: (
+                response: SubscriptionsResult,
+            ): SubscriptionListResponse => {
+                const subscriptions = response.subscriptions.page || [];
```
DutchBen, line 72: More of a general comment but I would map all the logic that converts the graphql result into a SubscriptionList Item (like applying mapGraphQlSubscriptionsResultToSubscriptionListItems) to this function. So you get items you can use directly from the hook

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionListSummary.ts
```diff
@@ -0,0 +1,70 @@
...
+            query: (variables) => ({
+                document: subscriptionListSummaryQuery,
+                variables,
+            }),
+            transformResponse: (
+                response: SubscriptionsResult<SubscriptionSummary>,
+            ): SubscriptionListSummaryResponse => {
+                const subscriptions = response.subscriptions.page || [];
+                const pageInfo = response.subscriptions.pageInfo || {};
+
```
DutchBen, line 58: Same comment as above

DutchBen (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionListSummary.ts
ricardovdheijden: It is hard to make a good abstraction here. The prop, in this case, `subscriptions` is different in most situations (processes, productBlocks etc)

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionList.ts
ricardovdheijden: I would not move such logic in here. Data from the backend should stay roughly in its shape. SubscriptionListItems are tailored for use in the table, typical view data. Adding that mapper here fully transforms the result from the graphql and makes possible reuse of this endpoint hard.

tjeerddie (approved):

PR #896, opened 2024-03-11
702 Moves subscription detail graphql query to RTK client

ricardovdheijden:
# 702 subscriptionDetail

todo:

- [x] Changeset


packages/orchestrator-ui-components/src/components/WfoForms/formFields/SubscriptionSummaryField.tsx
```diff
@@ -18,10 +18,9 @@ import { connectField, filterDOMProps } from 'uniforms';
 
 import { EuiFormRow, EuiText } from '@elastic/eui';
 
-import { GET_SUBSCRIPTION_DETAIL_GRAPHQL_QUERY } from '../../../graphqlQueries';
-import { useQueryWithGraphql } from '../../../hooks';
-import { WfoLoading } from '../../WfoLoading';
-import { WfoSubscriptionGeneral } from '../../WfoSubscription';
+import { WfoLoading, WfoSubscriptionGeneral } from '@/components';
+import { useGetSubscriptionDetailQuery } from '@/rtk/endpoints/subscriptionDetail';
```
DutchBen, line 22: I think it's better do import directly from @rtk

packages/orchestrator-ui-components/src/components/WfoForms/formFields/SubscriptionSummaryField.tsx
```diff
@@ -33,12 +32,10 @@ interface SubscriptionSummaryDisplayProps {
...
-    const { data } = useQueryWithGraphql(
-        GET_SUBSCRIPTION_DETAIL_GRAPHQL_QUERY,
-        { subscriptionId },
-        `subscription-${subscriptionId}`,
-    );
-    const subscriptionDetail = data?.subscriptions.page[0];
+    const { data } = useGetSubscriptionDetailQuery({
+        subscriptionId,
+    });
+    const subscriptionDetail = data?.subscriptions[0];
```
DutchBen, line 38: Isn't this always going to be 1 result. I would prefer moving this destucturing to the endpoint so you data is the subscriptionDetail right away

packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -15,8 +15,7 @@ export enum BaseQueryTypes {
 
 export enum CacheTags {
     engineStatus = 'engineStatus',
-    // This is a placeholder for now. Having only one messes up the type of the tag in the onCacheEntryAdded function.
-    testStatus = 'testStatus',
+    subscription = 'subscription',
```
DutchBen, line 18: Doesn't subscriptionDetail better describe the cache value?

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscription.tsx
```diff
@@ -47,18 +49,15 @@ export const WfoSubscription = ({ subscriptionId }: WfoSubscriptionProps) => {
...
+        subscriptionId,
+    });
 
     const onSelectedTabChanged = (tab: SubscriptionDetailTab) => {
         setActiveTab(tab);
     };
 
-    const subscriptionResult =
-        data && data.subscriptions && data.subscriptions.page;
+    const subscriptionResult = data && data.subscriptions;
```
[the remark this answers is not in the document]
ricardovdheijden: From the transform callback in rtk we always return the data like this. Looking at this piece of code i have the feeling we can get rid of the subscriptionResult const. Will have a look

packages/orchestrator-ui-components/src/rtk/api.ts
ricardovdheijden: I think it should be subscription. My thought: when setting the cachekey in `subscriptionDetail.ts` the subscriptionId is set as ID:
```
providesTags: (result, error, arg) => [
                {
                    type: CacheTags.subscription,
                    id: arg.subscriptionId,
                },
            ],
```
Invalidating this key would mean you will invalidate a specific subscription with id: yy

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscription.tsx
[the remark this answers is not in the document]
DutchBen: Yeah you are right. It's returned like this in other places aswell. Never mind

pboers1988 (approved):

PR #902, opened 2024-03-12
702 products and productsSummary

ricardovdheijden:
#702 
Products and ProductsSummary

PR #909, opened 2024-03-13
702 Moves subscriptions dropdown options to RTK

ricardovdheijden:
#702 
702 Moves subscriptions dropdown options to RTK

Extra's:
- Introduces a const defining "all items" to be fetched from a paginated api. This number is so large we can assume we will have all items


packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionsDropdownOptions.ts
```diff
@@ -0,0 +1,72 @@
...
+import {
+    GraphqlQueryVariables,
+    SubscriptionDropdownOption,
+    SubscriptionDropdownOptionsResult,
+} from '@/types';
+
+export const subscriptionsDropdownOptionsQuery = `
+    query SubscriptionDropdownOptions(
+        $filterBy: [GraphqlFilter!]
+        $first: Int!
```
DutchBen, line 12: I would hardcode the NUMBER_OF_ITEMS_REPRESENTING_ALL_ITEMS here instead of passing it in as a variable. It's always going to be there so why the extra step?

ricardovdheijden: I went for this extra step to have a definition of "infinite" or "all results" somewhere central in the app. In this current PR i am using the const on 1 place, but there is another usecase in:
- https://github.com/workfloworchestrator/orchestrator-ui-library/pull/902
- File `packages/orchestrator-ui-components/src/rtk/endpoints/startComboBoxOptions.ts` in PR #910

ricardovdheijden: It will make the query less readable and the value of subscriptionsDropdownOptionsQuery would not be valid graphql statement anymore. I actually would even prefer to move all these queries to actual `.graphql` files

Issue #638, opened 2024-01-11
Separate the SURF specific form fields from open source version


wouter1975:
After discussion, moving all surf specific fields to a 'deprecated' folder.

PR #915, opened 2024-03-14
638 move surf specific fields to deprecated folder

ricardovdheijden:
#638 Moves SURF specific fields to a deprecated folder


packages/orchestrator-ui-components/src/components/WfoForms/formFields/SubscriptionField.tsx
```diff
@@ -33,12 +33,13 @@ import {
     EuiText,
 } from '@elastic/eui';
 
-import { useOrchestratorTheme } from '../../../hooks';
-import { useGetSubscriptionDropdownOptions } from '../../../hooks/surf/useGetSubscriptionDropdownOptions';
-import { SubscriptionDropdownOption } from '../../../types';
+import { useOrchestratorTheme } from '@/hooks';
+import { useGetSubscriptionDropdownOptions } from '@/hooks/surf/useGetSubscriptionDropdownOptions';
```
DutchBen, line 37: Should we rename the hooks/surf folder to deprecated aswell? Also, maybe getting the subscriptionDropdown options isn't surf specific at all right?

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/SubscriptionField.tsx
ricardovdheijden: Let's discuss monday, sounds like something we should do.

ricardovdheijden: This seems to be the only surf specific part left, moving to deprecated folder

PR #918, opened 2024-03-14
Added isDisabled prop to WfoDropdownButton

Georgi2704:


DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoKeyValueTable/styles.ts
```diff
@@ -24,6 +24,8 @@ export const getStyles = (theme: EuiThemeComputed) => {
         rowNumber % 2 ? lightBackground : darkBackground;
 
     const keyColumnStyle = css({
+        display: 'flex',
+        alignItems: 'center',
```
ricardovdheijden, line 28: This causes trouble when you have multiple lines in a key-value table.
Before your adjustments:
image
After your adjustments:
image

What are you trying to fix here?

Georgi2704: I wanted this change so I can put the "Default Sending Level" and "Customers to Inform" inside the Service Tickets Detail Table: 

Before:
Screenshot 2024-03-18 at 14 07 28


Now:
Screenshot 2024-03-18 at 14 03 46

ricardovdheijden (approved):

Issue #102, opened 2023-06-13
Subscription detail support fetch external data - implement IPAM


acidjunk:
For this a federative graphql server is needed that can fetch external resource ID's (like ims_circuit_id for SURF)

A couple of fields in the V1 perform a resolve action when clicking the '+' icon. It will collect date from external systems like IPAM and CMDB.

In the V1 the solution offers:
* direct link to the external system using the `id`
* full details of the id.

Collapsed view of CMDB:

![Image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/83226590/59f05d15-0ad8-4341-8c8f-f35d6fb9580e)

Expanded view:
![Image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/83226590/f9f17ac4-b0ca-450d-870b-81d04b7d7e19)

Expanded view of IMS (no direct link to IPAM):
![Image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/83226590/5751be88-0919-4339-a757-86df3983be82)

API calls of V1:
* for IPAM: https://orchestrator.automation.surf.net/api/surf/ipam/prefix_by_id/6781 using GET
* for IMS: https://orchestrator.automation.surf.net/api/surf/ims/service_by_ims_service_id/54370 using GET

Start with a basic json payload presentation as used for the in use by subscription

```json
{
    "prefix": "145.145.0.54/31",
    "asn": 2,
    "state": 1,
    "description": "PtP for 118e2209-1e0b-4225-9877-17a179e82c94 a7b2d42a-5698-4b5c-81f9-b9e165c6e813",
    "vrf": 1,
    "tags": [
        "orchestrator",
        "ip"
    ],
    "id": 6780,
    "afi": "4",
    "asn__label": "AS0 (Default AS, do not delete)",
    "parent": 8,
    "parent__label": "default(4): 145.145.0.0/18",
    "state__label": "Default Resource: Ready",
    "vrf__label": "global",
    "addresses": [
        {
            "name": "e0-1-2-1111.ut042a-jnx-02.surf.net",
            "address": "145.145.0.54",
            "id": 13989,
            "description": null,
            "fqdn": "e0-1-2-1111.ut042a-jnx-02.surf.net",
            "mac": null,
            "prefix": 6780,
            "state": 1,
            "tags": [
                "orchestrator",
                "ip",
                "SURFnet"
            ],
            "vrf": 1,
[...]

PR #955, opened 2024-03-22
102 subscription detail custom field render

ricardovdheijden:
#102 

- Introduces a context provider and a hook to override values.
- The value gets overridden when it matches a given key. Overrides take place at the app side of the project.

Update 25-3: adding another reusable component
Update 26-3: rethinking the override mechanism considering next stories
Update 27-3: new solution in https://github.com/workfloworchestrator/orchestrator-ui-library/pull/972


packages/orchestrator-ui-components/src/contexts/ValueOverrideContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+        fieldValue: FieldValue,
+    ): React.ReactNode | null => {
+        if (!valueOverrideConfiguration) {
+            return null;
+        }
+
+        const renderFunctionForField: ValueOverrideFunction | undefined =
+            valueOverrideConfiguration[fieldValue.field];
+
+        // This check is needed TS does not infer the type correctly
```
DutchBen, line 39:
```suggestion
        // This check is needed because TS does not infer the type correctly
```

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock/WfoProductBlockKeyValueRow.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <>{value}</>
+        );
+
+    return (
+        <tr css={rowStyle}>
+            <td css={leftColumnStyle}>
+                <b>{camelToHuman(field)}</b>
+            </td>
+            <td css={rightColumnStyle}>
+                {valueOverride?.(fieldValue) ?? (
```
tjeerddie, line 40: I would like this as a const above the html, to have the logic outside the html

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock/WfoSubscriptionProductBlock.tsx
```diff
@@ -188,39 +160,10 @@ export const WfoSubscriptionProductBlock = ({
                                         ),
                                 )
                                 .map((productBlockInstanceValue, index) => (
-                                    <tr key={index}>
```
tjeerddie, line 191: not part of your PR but I would add the  `productBlockInstanceValues.filter()` above as a const outside the html

PR #972, opened 2024-03-27
102 subscription detail custom field render rtk

ricardovdheijden:
#102 

- introduces a config override mechanism using RTK
- the value gets overridden when it matches a given key. Overrides take place at the app side of the project.


packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock/WfoProductBlockKeyValueRow.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { EuiBadge } from '@elastic/eui';
+
+import { useSubscriptionDetailValueOverride } from '@/components';
+import { useWithOrchestratorTheme } from '@/hooks';
+import { FieldValue } from '@/types';
+import { camelToHuman } from '@/utils';
+
+import { getStyles } from './styles';
+
+export type WfoProductBlockKeyValueRowProps = {
```
DutchBen, line 12: No export needed here right?

packages/orchestrator-ui-components/src/rtk/slices/orchestratorComponentOverride.ts
```diff
@@ -0,0 +1,23 @@
+import { Slice, createSlice } from '@reduxjs/toolkit';
+
+import { ValueOverrideFunction } from '@/components/WfoSubscription/overrides/useSubscriptionDetailValueOverride';
```
DutchBen, line 3: I would move this type here instead and import it /useSubscriptionDetailValueOverride

packages/orchestrator-ui-components/src/rtk/slices/orchestratorComponentOverride.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export type ValueOverrideConfiguration = Record<string, ValueOverrideFunction>;
+
+export type OrchestratorComponentOverride = {
+    subscriptionDetail?: {
```
DutchBen, line 8: This is fine for now but I wonder if we get more overrides if you can't make this a bit more generic. So perphaps make this something like 
```{
  [string]: valueOverrides 
}```

Keep something like that in mind when adding more overrides

DutchBen (approved):

Georgi2704 (approved):

packages/orchestrator-ui-components/src/rtk/slices/orchestratorComponentOverride.ts
ricardovdheijden: Will take this in consideration for the future overrides (next stories)

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock/WfoProductBlockKeyValueRow.tsx
ricardovdheijden: Exporting the props type to provide override possibilities by clients

Issue #333, opened 2023-10-04
Add customer descriptions (aliases) and external links


wouter1975:
In V1 the customer descriptions (aliases) and external links are available, which is SURF specific features.

- [x] include the customer aliases in the subscription details table
- [x] external links can be added below the References section, where the metadata is displayed.

![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/57a4d9f9-3b8f-4056-894a-5abcb51b1e5e)

PR #980, opened 2024-04-02
333 subscription detail general section override

ricardovdheijden:
#333 

Adds an override for rendering the sections in the General tab on the SubscriptionDetail page. It exposes a function that provides all existing items and the current subscription object.


wouter1975 (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/customers.ts
```diff
@@ -3,25 +3,62 @@ import { Customer, CustomersResult } from '@/types';
...
-            transformResponse: (response: CustomersResult): Customer[] => {
-                const customers = response?.customers?.page || [];
-                return customers;
-            },
+            transformResponse: (response: CustomersResult): Customer[] =>
+                response?.customers?.page || [],
+        }),
+        getCustomer: build.query<
+            Customer[],
+            { customerIds: string | string[] }
```
tjeerddie, line 40: could always expect a array

tjeerddie (approved):

PR #1005, opened 2024-04-11
941: Add theme and websocket toggles to config

DutchBen:
Adds the feature toggles to the configuration so they can are used at runtime instead of buildtime


packages/orchestrator-ui-components/src/components/WfoPageTemplate/WfoPageHeader/WfoPageHeader.tsx
```diff
@@ -47,6 +46,7 @@ export const WfoPageHeader: FC<WfoPageHeaderProps> = ({
 }) => {
     const t = useTranslations('main');
     const { theme, multiplyByBaseUnit, colorMode } = useOrchestratorTheme();
+    const orchestratorConfig = useGetOrchestratorConfig();
```
ricardovdheijden, line 49: You could destructure here directly

ricardovdheijden (approved):

Georgi2704 (approved):

Issue #1031, opened 2024-04-18
Improve resolving of ' In-use by subscription(s)'


DutchBen:

Georgi2704:
Also discussed that about the productBlockInstances with @mrijk 
similarly. issue:https://github.com/workfloworchestrator/orchestrator-ui-library/issues/782

There could be a resolver for the Owner Subscription Description for a product block, not sure if it should be part of this specific issue.

PR #1032, opened 2024-04-18
776 in use by relations

DutchBen:
Replaces the jsonCode block output for the inUse by subscrtiptiosn field of the product block instance cards on the subscriptionDetail page

This is a possible follow up ticket that moves this functionality back to the backend: https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1031

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoInUseByRelations.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoInUseByRelations.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { useTranslations } from 'next-intl';
+import Link from 'next/link';
 
 import {
     WfoError,
@@ -9,8 +10,11 @@ import {
     WfoLoading,
 } from '@/components';
 import type { WfoKeyValueTableDataType } from '@/components';
+import { useWithOrchestratorTheme } from '@/hooks';
 import { useGetInUseByRelationDetailsQuery } from '@/rtk';
 import { InUseByRelation, InUseByRelationDetail } from '@/types';
+
+import { getStyles } from './styles';
 
 interface WfoInUseByRelationsProps {
     inUseByRelations: InUseByRelation[];
@@ -20,7 +24,7 @@ export const WfoInUseByRelations = ({
     inUseByRelations,
 }: WfoInUseByRelationsProps) => {
     const t = useTranslations('subscriptions.detail');
-
+    const { inUseByRelationDetailsStyle } = useWithOrchestratorTheme(getStyles);
     const subscriptionIds = inUseByRelations
         .map((relation) => relation.subscription_id)
         .join('|');
@@ -41,16 +45,6 @@ export const WfoInUseByRelations = ({
     ): WfoKeyValueTableDataType[] => {
         return [
             {
-                key: t('description'),
-                value: inUseByRelationDetails.description,
-                textToCopy: inUseByRelationDetails.description,
-            },
-            {
-                key: t('productType'),
-                value: inUseByRelationDetails.product.productType,
-                textToCopy: inUseByRelationDetails.product.productType,
-            },
-            {
                 key: t('subscriptionId'),
                 value: (
                     <WfoFirstPartUUID
@@ -58,6 +52,22 @@ export const WfoInUseByRelations = ({
                     />
                 ),
                 textToCopy: inUseByRelationDetails.subscriptionId,
+            },
+            {
+                key: t('description'),
+                value: (
+                    <Link
+                        href={`/subscriptions/${inUseByRelationDetails.subscriptionId}`}
+                    >
+                        {inUseByRelationDetails.description}
+                    </Link>
+                ),
+                textToCopy: inUseByRelationDetails.description,
+            },
+            {
+                key: t('productType'),
+                value: inUseByRelationDetails.product.productType,
+                textToCopy: inUseByRelationDetails.product.productType,
             },
         ];
     };
@@ -68,11 +78,12 @@ export const WfoInUseByRelations = ({
                 const keyValues = getKeyValues(relation);
 
                 return (
-                    <WfoKeyValueTable
-                        keyValues={keyValues}
-                        showCopyToClipboardIcon={true}
-                        key={index}
-                    />
+                    <div css={inUseByRelationDetailsStyle} key={index}>
+                        <WfoKeyValueTable
+                            keyValues={keyValues}
+                            showCopyToClipboardIcon={true}
+                        />
+                    </div>
                 );
             })}
         </>
--- a/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoInUseByRelations.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSubscription/WfoInUseByRelations.tsx
@@ -1,8 +1,16 @@
 import React from 'react';
 
-import { WfoError, WfoLoading } from '@/components';
+import { useTranslations } from 'next-intl';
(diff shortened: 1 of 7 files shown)


packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionInUseByRelationsList.ts
```diff
@@ -0,0 +1,54 @@
+import { CacheTags, orchestratorApi } from '@/rtk';
+import { InUseByRelationDetail, InUseByRelationsDetailResult } from '@/types';
+
+export const subscriptionInUseByRelationQuery = `
+    query SubscriptionsInUseByRelationsDetails(
+        $subscriptionIds: String!
+    ) {
+        subscriptions(
+            first: 1000000
```
mrijk, line 9: I don't think this is the correct way to get all in use by subscriptions. It's paginated, so probably easiest is to write some function that iterates over all pages and yields the results.

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionInUseByRelationsList.ts
```diff
@@ -0,0 +1,54 @@
...
+            after: 0
+            filterBy: [{field: "subscriptionId", value: $subscriptionIds},{field: "status", value: "INITIAL|ACTIVE|MIGRATING|DISABLED|PROVISIONING"}]
```
mrijk, line 11: Maybe collect all magic numbers/strings in some settings?

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionInUseByRelationsList.ts
```diff
@@ -0,0 +1,54 @@
+import { CacheTags, orchestratorApi } from '@/rtk';
+import { InUseByRelationDetail, InUseByRelationsDetailResult } from '@/types';
+
+export const subscriptionInUseByRelationQuery = `
+    query SubscriptionsInUseByRelationsDetails(
```
mrijk, line 5: As discussed briefly yesterday, I'm a bit fearful that we are going to end up with using Graphql as REST. If possible, we should limit the amount of queries and use resolvers instead of having multiple queries to build a single page.

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoInUseByRelations.tsx
```diff
@@ -0,0 +1,91 @@
...
+    inUseByRelations: InUseByRelation[];
+}
+
+export const WfoInUseByRelations = ({
+    inUseByRelations,
+}: WfoInUseByRelationsProps) => {
+    const t = useTranslations('subscriptions.detail');
+    const { inUseByRelationDetailsStyle } = useWithOrchestratorTheme(getStyles);
+    const subscriptionIds = inUseByRelations
+        .map((relation) => relation.subscription_id)
```
mrijk, line 29: In the past I used lodash (and similar libraries) to get values from objects and avoid having to endlessly write mapping function.

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionInUseByRelationsList.ts
```diff
@@ -0,0 +1,54 @@
...
+        getInUseByRelationDetails: builder.query<
+            InUseByRelationsDetailResponse,
+            { subscriptionIds: string }
+        >({
+            query: (variables) => ({
+                document: subscriptionInUseByRelationQuery,
+                variables,
+            }),
+            transformResponse: (
+                response: InUseByRelationsDetailResult,
```
mrijk, line 40: I still think this defeats the purpose of graphql, adding types to each and every query result since the whole point is being flexible. Feel free to ignore my ranting about typing in front-ends ;)

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionInUseByRelationsList.ts
DutchBen: We have had this discussion before and I don't think it's the right way to get all results by doing extra queries just to make sure there aren't any extra results. That means extra unneeded http requests.

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionInUseByRelationsList.ts
DutchBen: There is a ticket for that: https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1031

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionInUseByRelationsList.ts
mrijk: You can have both. A collector with a default fetch size of for example 1000 (or any other reasonable number), that gets all your items in batches of 1000. And yes, there will be additional requests if you have 1001 items. And the possibility to override the default batch size if you really want to get data in larger chunks. My point is mainly that you want to abstract away from pages in these cases, and avoid having magic numbers all over the place.

mrijk (approved):
LGTM

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoInUseByRelations.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    if (isError) {
+        return <WfoError />;
+    }
+
+    if (isLoading) {
+        return <WfoLoading />;
+    }
+
+    const getKeyValues = (
```
ricardovdheijden, line 43: isn't this more a mapper?

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoInUseByRelations.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface WfoInUseByRelationsProps {
+    inUseByRelations: InUseByRelation[];
+}
+
+export const WfoInUseByRelations = ({
+    inUseByRelations,
+}: WfoInUseByRelationsProps) => {
+    const t = useTranslations('subscriptions.detail');
+    const { inUseByRelationDetailsStyle } = useWithOrchestratorTheme(getStyles);
+    const subscriptionIds = inUseByRelations
```
ricardovdheijden, line 28: This is input for the query to the backend, it might be better to move the mapper with `|` to the data fetch part of code. In the end we just want to provide a list (array) of id's to fetch data for

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionInUseByRelationsList.ts
DutchBen: Replaced the number with the const. To be complete. We have in fact implemented paging as you described for pages that actually have paging. For pages that just need everything the results of discussions with backenders have been to not implement something for that so all we have at the frontend is using a large number or making unneeded requests just to satisfy some code esthetics. My choice here is to make what we do explicit and not do the unneeded calls

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoInUseByRelations.tsx
DutchBen: There is a ticket to improve this and resolve all of it in the backend: https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1031

Issue #780, opened 2024-02-14
[bug] L3VPN shows wrong VLAN validator message


wouter1975:
V2:
![Image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/83226590/509d3084-03ef-48c0-8573-ec218350b422)

V1:

![Image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/83226590/4522b75c-8dd9-4400-8f5a-d5e18a638d5e)

DutchBen:
@wouter1975 The message shown is different from V1 but isn't the real fix to not show a message at all until a post is selected?

wouter1975:
@DutchBen , that sounds like a good solution indeed. I recall that we might need to show the message on the vlan field in case of untagged ports or ports  used in a core_link, can we keep this, whie we hide it in the situation of a L3VPN?

PR #1043, opened 2024-04-22
780 vlan validator message

DutchBen:
https://github.com/workfloworchestrator/orchestrator-ui-library/issues/780 
The tickets describes shows the message that shows is different than in V1. The real fix IMO and the one implemented in this PR is to not show a message in a all until a port subscription is selected.

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoForms/formFields/ListAddField.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoForms/formFields/ListAddField.tsx
@@ -53,9 +53,9 @@ function ListAdd({
         !disabled &&
         !(
             parent.maxCount! <=
-            Math.max(initialCount ?? 0, parent.value!.length)
+            Math.max(initialCount ?? 0, parent?.value!.length)
         );
-    const count = 1 + Math.max((initialCount ?? 0) - parent.value!.length, 0);
+    const count = 1 + Math.max((initialCount ?? 0) - parent?.value!.length, 0);
 
     function onAction(event: React.KeyboardEvent | React.MouseEvent) {
         if (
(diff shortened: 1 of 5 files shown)


packages/orchestrator-ui-components/src/components/WfoForms/formFields/ListAddField.tsx
```diff
@@ -53,9 +53,9 @@ function ListAdd({
         !disabled &&
         !(
             parent.maxCount! <=
-            Math.max(initialCount ?? 0, parent.value!.length)
+            Math.max(initialCount ?? 0, parent?.value!.length)
         );
-    const count = 1 + Math.max((initialCount ?? 0) - parent.value!.length, 0);
+    const count = 1 + Math.max((initialCount ?? 0) - parent?.value!.length, 0);
```
mrijk, line 58: Could maybe be combined with a single assignment to a var and resused in 56?

packages/orchestrator-ui-components/src/components/WfoForms/formFields/deprecated/VlanField.tsx
```diff
@@ -227,10 +227,8 @@ function Vlan({
             : undefined;
 
     let message = '';
-    if (!imsIsLoading) {
-        if (missingInIms) {
-            message = t('widgets.vlan.missingInIms');
-        } else if (portIsTagged && nsiVlansOnly) {
+    if (!imsIsLoading && subscriptionId) {
+        if (portIsTagged && nsiVlansOnly) {
```
mrijk, line 231: Can I just ignore changes to these formfields since they are in deprecated? If not, why is the frontend aware of nsivlans???

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/deprecated/VlanField.tsx
DutchBen: Yes just ignore them

packages/orchestrator-ui-components/src/components/WfoForms/formFields/ListAddField.tsx
DutchBen: Doesn't do the same thing... I think

Georgi2704 (approved):

Issue #367, opened 2023-10-16
Make product domain boundary visible in subscription detail page


wouter1975:
With an unlimited coupling of product blocks, it is not clear where the product domain model has ended. It would be good to distinguish which resource types and product blocks belong to the product definition, and which fall outside the product domain boundary.

The mechanism to make this distinguish is based on the `owner_subscription_id`.
* if the `owner_subscription_id` is not the same as the owner of the current subscription page, than the next product block in the outside the product domain model.
* If `parent product block` subscription owner == owner of the current subscription page, that product block is part of subscription.

### Visual presentation of the boundary
* The product blocks not part of the subscription should be reflected in a less active state (eg grey text in product bock tree)
  * in the tree view make the product block title less bold 
  * Product block card: background lighter and maybe font colour gray?
* Example shall be prepared after completion of #282

wouter1975:
Discuss whether it would be good to hide all resource types of a product block of another subscription owner. And only show the owner subscription with a link. Clicking on the show details shows all information of the PB.

![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/25c2539c-e111-41fb-9931-7f1047cd8632)

This would be especially good for IP_prefix, which the parent product block is a repetition of the productblock.



![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/1df7d437-8d62-46df-96c7-eebd9e444d45)

Will become:



![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/a03820f3-3a68-4619-a6d1-f8d6ec3667c1)

PR #1048, opened 2024-04-24
367 subscription productblocks outside the current subscription

DutchBen:
This ticket has to wait for https://github.com/workfloworchestrator/orchestrator-core/issues/626
Implements https://github.com/workfloworchestrator/orchestrator-ui-library/issues/367

Adds a different style if the productBlock instance is outside of the current subscription

### DIFF
--- a/packages/orchestrator-ui-components/src/types/types.ts
+++ b/packages/orchestrator-ui-components/src/types/types.ts
@@ -92,7 +92,7 @@ export interface TreeBlock extends ProductBlockInstance {
     label: string | number | boolean;
     callback: () => void;
     children: TreeBlock[];
-    outsideCurrentSubscription: boolean;
+    isOutsideCurrentSubscription: boolean;
 }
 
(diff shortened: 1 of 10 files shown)


packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -92,6 +92,7 @@ export interface TreeBlock extends ProductBlockInstance {
     label: string | number | boolean;
     callback: () => void;
     children: TreeBlock[];
+    outsideCurrentSubscription: boolean;
```
ricardovdheijden, line 95: Suggesting naming this `isOutsideCurrentSubscription`

ricardovdheijden (approved):

wouter1975 (approved):

packages/orchestrator-ui-components/src/types/types.ts
DutchBen: Done

PR #1107, opened 2024-05-14
1086 total number of products fix

### LINKED ISSUE #1086: Total number of products should not include terminated subscriptions

ricardovdheijden:
#1086 

Adds extra filter to the query used in the summary card showing the total number of products

### DIFF
--- a/packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
+++ b/packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
@@ -100,12 +100,19 @@ export const taskListSummaryQueryVariables: GraphqlQueryVariables<Process> = {
     ],
 };
 
-export const productsSummaryQueryVariables: GraphqlQueryVariables<ProductsSummary> =
-    {
-        first: 1000,
-        after: 0,
-        sortBy: {
-            field: 'name',
-            order: SortOrder.ASC,
+export const productsSummaryQueryVariables: GraphqlQueryVariables<
+    ProductsSummary & Subscription
+> = {
+    first: 1000,
+    after: 0,
+    sortBy: {
+        field: 'name',
+        order: SortOrder.ASC,
+    },
+    filterBy: [
+        {
+            field: 'status',
+            value: '!terminated',
         },
-    };
+    ],
+};
--- a/packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
+++ b/packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
@@ -1,3 +1,4 @@
+import { MAXIMUM_ITEMS_FOR_BULK_FETCHING } from '@/configuration';
 import {
     GraphqlFilter,
     GraphqlQueryVariables,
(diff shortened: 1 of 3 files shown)


packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
```diff
@@ -100,12 +101,19 @@ export const taskListSummaryQueryVariables: GraphqlQueryVariables<Process> = {
...
+    first: 1000,
+    after: 0,
+    sortBy: {
+        field: 'name',
+        order: SortOrder.ASC,
+    },
+    filterBy: [
+        {
+            field: 'status',
+            value: `!${SubscriptionStatus.TERMINATED}`,
```
mrijk, line 116: I can imagine this filter happens quite a lot. Maybe have a couple of predefined field/value pairs?

packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
```diff
@@ -100,12 +101,19 @@ export const taskListSummaryQueryVariables: GraphqlQueryVariables<Process> = {
...
-    {
-        first: 1000,
-        after: 0,
-        sortBy: {
-            field: 'name',
-            order: SortOrder.ASC,
+export const productsSummaryQueryVariables: GraphqlQueryVariables<
+    ProductsSummary & Subscription
+> = {
+    first: 1000,
```
mrijk, line 107: Magic number.

packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
```diff
@@ -100,12 +101,19 @@ export const taskListSummaryQueryVariables: GraphqlQueryVariables<Process> = {
...
 
-export const productsSummaryQueryVariables: GraphqlQueryVariables<ProductsSummary> =
-    {
-        first: 1000,
-        after: 0,
-        sortBy: {
-            field: 'name',
-            order: SortOrder.ASC,
+export const productsSummaryQueryVariables: GraphqlQueryVariables<
+    ProductsSummary & Subscription
```
mrijk, line 105: I'm not very familiar with the typescript type system, but having a union type of a ProductSummary and a Subscription feels a bit weird, since they are kind of unrelated. Unless this is just meant as a quick way to avoid introducing 'real' types. In that case you might be better of with something like (if that is even possible): 'type myCombinedWhateverType = ProductsSummary & Subscription' and use that

ricardovdheijden: In this query both are actually related: counting the number of subscriptions per product

packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
ricardovdheijden: Indeed, when also used somewhere else we need to start introducing predefined filters. This is currently not the case

packages/orchestrator-ui-components/src/pages/startPage/queryVariables.ts
ricardovdheijden: Replaced with a constant

mrijk (approved):
LGTM

DutchBen (approved):

PR #1127, opened 2024-05-22
1126 signout error

DutchBen:


packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -87,7 +87,11 @@ export const orchestratorApi = createApi({
                     prepareHeaders,
                     customErrors: (error) => {
                         const { name, message, stack, response } = error;
-                        if (response?.errors && response.errors?.length > 0) {
+                        if (
+                            response?.errors &&
```
ricardovdheijden, line 91: `response?.errors && response.errors?.length > 0` can be replaced with `response?.errors?.length > 0` for better readability.

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/rtk/api.ts
tjeerddie: you don't even need the `> 0`

tjeerddie (approved):

packages/orchestrator-ui-components/src/rtk/api.ts
DutchBen: Can you do a 'code suggestion' for these types of comments next time?

PR #1139, opened 2024-05-28
1124 port select bug

DutchBen:
This PR adds a notice when - in a modify workflow - a previously selected subscription with a node doesn't exist anymore.

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoForms/formFields/deprecated/ImsPortIdField.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoForms/formFields/deprecated/ImsPortIdField.tsx
@@ -157,65 +157,71 @@ function ImsPortId({
         }))
         .sort((x, y) => x.label.localeCompare(y.label));
     const portValue = portOptions.find((option) => option.value === value);
-
     return (
         <EuiFlexItem css={imsPortIdFieldStyling}>
             <section {...filterDOMProps(props)}>
                 <EuiFormRow
                     label={label}
-                    labelAppend={<EuiText size="m">{description}</EuiText>}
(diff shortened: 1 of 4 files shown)


packages/orchestrator-ui-components/src/components/WfoForms/formFields/deprecated/ImsPortIdField.tsx
```diff
@@ -157,65 +157,71 @@ function ImsPortId({
...
-                                        portOptions.length === 0
-                                    }
-                                    styles={reactSelectInnerComponentStyles}
-                                />
-                            </EuiFormRow>
-                        </div>
-                    </section>
+                    {(!nodeValue && nodeSubscriptionId && (
+                        <EuiText>
+                            This Node subscription is missing. Please check the
```
Georgi2704, line 173: Needs translation here

Georgi2704 (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/deprecated/ImsPortIdField.tsx
DutchBen: Well... yes and no. The field is surf specific and is set to be deprecated ie moved to surf specific. The translations currently available are the ones that are provided by the backend so there would need to be a translation added to the default translations file that would need to be removed ones the field is moved. So all and all beter to leave it untranslated for now I think

ricardovdheijden: This is a bit of a grey area, normally i would say that all user-facing texts should be translated using our translation mechanism.

ricardovdheijden (approved):

Issue #1108, opened 2024-05-15
Don't allow optional ENV vars


ricardovdheijden:
Currently we have several env vars that are optional to override default values. In certain cases, like `NEXTAUTH_WELL_KNOWN_OVERRIDE`, this ends up in different behavior of the app.

This might lead to confusion. In this ticket we need to find a solution for this. Possible solutions could be:
- Not having optional environment variables (throw errors)
- Listing the missing environment variables in the console of the app server (server side logging)

Conclusions of topics above
- [ ] Enforce all env variables
- [ ] Use same order at all place/yaml files etc
- [ ] Add documentation on workfloworchestrator.org
- [ ] Try to prevent to call `getEnvironmentVariables` in browser (currently, suppressing the warning)

PR #1149, opened 2024-06-03
1108 fetch env vars with warnings

ricardovdheijden:
#1108 

- Updates the type of context providers to be possibly `null`. This is more realistic since the value is not available when the code runs in the browser context.
- Adds a helper function to retrieve multiple environment variables. The `getEnvironmentVariables` function aggregates all missing variables and logs a warning in the console if anything is missing.

### DIFF
--- a/packages/orchestrator-ui-components/src/utils/getEnvironmentVariables.spec.ts
+++ b/packages/orchestrator-ui-components/src/utils/getEnvironmentVariables.spec.ts
@@ -20,7 +20,7 @@ describe('getEnvironmentVariables()', () => {
         process.env = originalEnv;
     });
 
-    it('returns an empty string as value in the result when the environment variable is not set', () => {
+    it('returns an object with the environment variables and its values', () => {
         process.env.test01 = 'value01';
         process.env.test02 = 'value02';
 
@@ -33,7 +33,7 @@ describe('getEnvironmentVariables()', () => {
         });
     });
 
-    it('returns an object with the environment variables and its values', () => {
+    it('returns an empty string as value in the result when the environment variable is not set', () => {
         process.env.test01 = 'value01';
         process.env.test03 = 'value03';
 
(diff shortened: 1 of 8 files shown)


packages/orchestrator-ui-components/src/utils/getEnvironmentVariables.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+            ...originalEnv,
+        };
+        jest.resetAllMocks();
+    });
+
+    afterEach(() => {
+        process.env = originalEnv;
+    });
+
+    it('returns an empty string as value in the result when the environment variable is not set', () => {
```
DutchBen, line 23: This text doesn't seem to match what is tested

DutchBen (approved):

packages/orchestrator-ui-components/src/utils/getEnvironmentVariables.spec.ts
ricardovdheijden: Done

packages/orchestrator-ui-components/src/utils/getEnvironmentVariables.ts
```diff
@@ -0,0 +1,33 @@
+import process from 'process';
+
+// By convention, only this function should be used to access the process.env object.
+// It trows an error if any of variables is not set
```
Georgi2704, line 4:
```suggestion
// It throws an error if any of variables is not set
```

Georgi2704 (approved):

packages/orchestrator-ui-components/src/utils/getEnvironmentVariables.ts
ricardovdheijden: This reminds me that that comment is a bit outdated. Thanks!

PR #1179, opened 2024-06-10
1178: Rethrows form validation errors in stepform so they can be hand…

DutchBen:
…led instead of supressed

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoStep/WfoStepForm.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoStep/WfoStepForm.tsx
@@ -2,8 +2,9 @@ import React, { useState } from 'react';
 
 import { EuiFlexItem } from '@elastic/eui';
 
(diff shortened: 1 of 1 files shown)


packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoStep/WfoStepForm.tsx
```diff
@@ -27,16 +29,28 @@ export const WfoStepForm = ({
...
+        return resumeProcess({ processId, userInputs: processInput })
+            .unwrap()
+            .then(() => {
                 setIsProcessing(true);
-            },
-        );
+            })
+            .catch((error) => {
+                if (error?.status !== HttpStatus.FormNotComplete) {
+                    if (error?.status === HttpStatus.BadRequest) {
```
ricardovdheijden, line 39: If you put this predicated as negated in the if above (line 38) you would end up with the same functionality, right?

ricardovdheijden (approved):

PR #1188, opened 2024-06-13
1163 update cache invalidation messages

DutchBen:


packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -562,3 +562,14 @@ export type SubscriptionActions = {
     terminate: SubscriptionAction[];
     system: SubscriptionAction[];
 };
+
+export type CacheTag = { type: CacheTagTypes; id?: string };
+
+export enum CacheTagTypes {
```
ricardovdheijden, line 568: Enums shouldn't be plural: `CacheTagTypes` -> `CacheTagType`

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
```diff
@@ -68,53 +67,25 @@ const streamMessagesApi = orchestratorApi.injectEndpoints({
...
-                        const cacheTags: CacheInvalidationTag[] = [cacheTag];
-
-                        if (id) {
-                            cacheTags.push({
-                                type: cacheTag,
-                                id,
-                            });
+                const invalidateTag = (cacheTag: CacheTag) => {
+                    const tagType = (() => {
+                        if (typeof cacheTag === 'object') {
```
ricardovdheijden, line 72: According to the type definition in types.ts it is always an object, therefore this check is redundant.

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/types/types.ts
DutchBen: Fixed

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
DutchBen: fixed

Georgi2704 (approved):

Issue #1164, opened 2024-06-05
Improve data reloading after cache invalidation


DutchBen:
When we receive a cache invalidation message through the websocket the affected data gets reloaded. When you are on a page that show this data it shows you a loading state when the data is being reloaded. Pages where this happens include the workflows listings page and the subscriptions listings page. 

We should consider that on some pages we might want to handle data updates diffently than others. On some pages the data can update 'silently' while on others - like a subscription detail - you might want to let a user know things are changing.

PR #1192, opened 2024-06-14
1164: Make a distinction between first data load and subsequent ones

DutchBen:
Ticket: https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1164

This causes a cache invalidation to show the blue night rider effect but not the '..loading' state.


Georgi2704 (approved):

ricardovdheijden:
Note: did a temporary adjustment (https://github.com/workfloworchestrator/orchestrator-ui-library/pull/1219) while waiting for the conclusion of the discussion in this PR

packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -572,4 +572,9 @@ export enum CacheTagType {
     subscriptions = 'subscriptions',
 }
 
+export type LoadingState = {
+    isLoading?: boolean;
+    isFetching?: boolean;
+};
```
ricardovdheijden, line 578: I think an enum would be better since you can specifically name the states:
`Loading`, `RefreshingData`, `Done`. Also the 2 booleans can represent 4 states, but i can only name 3. This means that one combination of the booleans isLoading and isFetching is invalid.

DutchBen: The main argument I have against this is that when checking the value now you can do ```
```if(loadingState.isLoading)``` and with an enum you would have to do '''loadingState === LoadingState.Loading``` In any case, I think this is really detailed and opinionated stuff and a bit much to hold up this issue. I prefer to merge it now.

Issue #1169, opened 2024-06-05
[NMS] Create chassis detail page - Services tab


wouter1975:
Implement https://git.ia.surfsara.nl/netdev/automation/projects/orchestrator-ui/-/issues/3

* ~Implement the services table based on dynamic data, which requires an intelligent recursive depends-on function, to include all underlaying subscription, but must stop (but including) at the following product types: LightPath/L2VPN/L3VPN/FW/IP/IPP, and show all possible subscriptions which are product type of `Port`.~ This function needs to be provided by the Backend, new ticket to be defined.
* Implement the services table based on depends-on relation to the port subscriptions of the current chassis, ensure the result returns a unique list.
* For each product type present on the node, it will be sorted and counted, by expanding the rows the actual services are shown.
* use the product type for the service column
* Breadcrumb uses the uuid insteed of the node name.

![Image](https://github.com/workfloworchestrator/orchestrator-ui-library/assets/83226590/508c3576-92e8-49de-be48-085c0b22c5f4)

PR #1218, opened 2024-06-19
1169 grouped table

ricardovdheijden:
#1169 

Creates table component that accepts grouped data.

image


ricardovdheijden:
Moving to internal repository until it is feature complete

ricardovdheijden:
Moves logic to internal repo

PR #1219, opened 2024-06-20
Workflows tasks list loading state

ricardovdheijden:
While waiting for the conclusion of discussion in PR https://github.com/workfloworchestrator/orchestrator-ui-library/pull/1192
processing this quick-fix in the current PR to prevent glitchy behavior when data gets refreshed in Task list and Workfow list pages

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
@@ -244,13 +244,9 @@ export const WfoProcessesList = ({
     return (
         <WfoTableWithFilter<ProcessListItem>
             queryString={queryString}
-            data={
-                !isFetching
-                    ? mapGraphQlProcessListResultToProcessListItems(
-                          processes || [],
-                      )
-                    : []
-            }
+            data={mapGraphQlProcessListResultToProcessListItems(
+                processes || [],
+            )}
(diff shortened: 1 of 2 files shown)


packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
```diff
@@ -244,13 +244,9 @@ export const WfoProcessesList = ({
...
-            data={
-                !isFetching
-                    ? mapGraphQlProcessListResultToProcessListItems(
-                          processes || [],
-                      )
-                    : []
-            }
+            data={mapGraphQlProcessListResultToProcessListItems(
+                processes || [],
+            )}
```
Mark90, line 249: `WfoSubscriptionsList` does it slightly different, not sure if that matters?

```
data={
                data
                    ? mapGraphQlSubscriptionsResultToSubscriptionListItems(data)
                    : []
            }
```

Mark90 (approved):

packages/orchestrator-ui-components/src/components/WfoProcessList/WfoProcessesList.tsx
ricardovdheijden: Effectively it is the same, but will update that one too

Georgi2704 (approved):

PR #1231, opened 2024-06-25
Fix not logging out when token expires

Georgi2704:


packages/orchestrator-ui-components/src/components/WfoTable/utils/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const DEFAULT_PAGE_SIZES = [5, 10, 15, 20, 25, 50, 100];
+export const DEFAULT_PAGE_SIZES = [5, 10, 15, 20, 25, 50, 100, 0];
```
Georgi2704, line 1: Also add "Show all rows"

packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -48,6 +53,24 @@ export const handlePromiseErrorWithCallback = <T>(
     });
 };
 
+export const catchErrorResponse: ResponseHandler = async (
+    response: Response,
+) => {
+    if (
+        response.status < HttpStatus.Ok ||
+        response.status >= HttpStatus.MultipleChoices
```
ricardovdheijden, line 61: 1st: this predicate is hard to follow ( "smaller than OK" or gte MultipleChoices ) -- consider extracting a predicate that is human readable, like "isNotSuccefsfulResponse".
2nd: in case of an error 401-unauthorized of 403-forbidden, wouldnt that trigger this if-case too? and thus never falling into the else-if and signing out the user?

packages/orchestrator-ui-components/src/rtk/api.ts
```diff
@@ -48,6 +53,24 @@ export const handlePromiseErrorWithCallback = <T>(
...
+    ) {
+        console.error(response.status, response.body);
+    } else if (
+        response.status === HttpStatus.Unauthorized ||
+        response.status === HttpStatus.Forbidden
```
ricardovdheijden, line 66: Same as above: consider a human readable predicate.

packages/orchestrator-ui-components/src/rtk/api.ts
Georgi2704: 2nd: true, nice catch

packages/orchestrator-ui-components/src/rtk/api.ts
Georgi2704: @ricardovdheijden so a seperate const ? also for the 1st one ?

const isUnauthorized = (status: number) => (status === HttpStatus.Unauthorized || status === HttpStatus.Forbidden)

DutchBen (approved):

ricardovdheijden (approved):

PR #1267, opened 2024-07-15
Fix 1239 text in form not visible

Georgi2704:
Before:
343169771-768893a5-e623-4a4c-9b61-cce33b1ad86c

After:
Screenshot 2024-07-15 at 16 52 36

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoForms/formFields/SummaryField.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoForms/formFields/SummaryField.tsx
@@ -19,6 +19,7 @@ import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 import { EuiFlexItem, EuiFormRow, EuiText } from '@elastic/eui';
 
 import { getStyles } from '@/components/WfoForms/formFields/SummaryFieldStyling';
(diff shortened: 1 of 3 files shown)


DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/SummaryField.tsx
```diff
@@ -19,6 +19,7 @@ import { FieldProps, connectField, filterDOMProps } from 'uniforms';
 import { EuiFlexItem, EuiFormRow, EuiText } from '@elastic/eui';
 
 import { getStyles } from '@/components/WfoForms/formFields/SummaryFieldStyling';
+import { getCommonFormFieldStyles } from '@/components/WfoForms/formFields/commonStyles';
```
ricardovdheijden, line 22: Import can be shorter

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/SummaryField.tsx
Georgi2704: I think we should fix them in all places then, because on all the other FormFields `getCommonFormFieldStyles `long import is used. IMO if I shorten this one it will make it more confusing.

PR #1279, opened 2024-07-23
1217 list field

DutchBen:
To startof any discussions: 
- I have taken a lot of liberty and css hacks here but it's because it's uniforms and doing it more clean would take a lot more time and introduces uncertainty. 
- I want Floris to test this well because i'm not fully aware off all scenarios. Changes may be incoming based on what he finds I guess.

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoForms/formFields/ListDelField.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoForms/formFields/ListDelField.tsx
@@ -49,14 +49,14 @@ function ListDel({
         { absoluteName: true },
     )[0];
 
+    const parentValueCount =
+        parent && parent.value && parent.value.length ? parent.value.length : 0;
+
     // Make a row deletable only if its:
-    // - doing so doesn't go below the minCount for the field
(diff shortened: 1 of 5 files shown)


packages/orchestrator-ui-components/src/components/WfoForms/formFields/ListDelField.tsx
```diff
@@ -49,20 +49,19 @@ function ListDel({
...
+    const isDeletable: boolean =
         !disabled &&
-        parent.minCount &&
-        parent.value &&
-        !(parent.minCount >= parent.value.length);
+        !readOnly &&
+        (!parent.minCount ||
+            (parent.value && parent.value.length >= parent.minCount))
+            ? true
+            : false;
```
ricardovdheijden, line 61: Removing these 2 lines should end up in the same result

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/ListDelField.tsx
DutchBen: This ensures a possible result is not undefined

Issue #1312, opened 2024-08-05
Ports are being cached and not re-retrieved


wouter1975:
See https://git.ia.surfsara.nl/netdev/automation/projects/orchestrator-ui/-/issues/53

PR #1341, opened 2024-08-14
1312 Enables refetching on mount to keep the ports list up to date

ricardovdheijden:
#1312

Enables refetching on mount to keep the ports list up to date


Georgi2704 (approved):

DutchBen:
Are you sure this fixes the problem? From the very brief issue description I suspect its an issue where a trigger through a websocket is needed. After this fix a stale page will still be stale until it remounts: https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1312

DutchBen (approved):

PR #1375, opened 2024-08-27
822: Reset apiState and reconnect websocket on return of focus

DutchBen:


packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
```diff
@@ -23,12 +24,11 @@ const getWebSocket = async (url: string) => {
     }
 };
 
-// Recommended values are between 30 and 60 seconds
 // Note: When the browser tab is suspended, the browser will give this suspended tab approximately every 60 seconds time
 // to execute code. This means that the ping message will be sent every 60 seconds in suspended mode. This depends
 // on the browser and can differ.
 const PING_INTERVAL_MS = 45000;
-
+const DEBOUNCE_CLOSE_INTERVAL_MS = 60000;
```
ricardovdheijden, line 31: Think > 60 might be better, we might risk disconnects from FE. Maybe 70?

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
```diff
@@ -88,8 +88,15 @@ const streamMessagesApi = orchestratorApi.injectEndpoints({
                 // Send a ping message every to the websocket server to keep the connection alive
                 const pingInterval = setInterval(() => {
                     webSocket.send('__ping__');
```
ricardovdheijden, line 90: A very nice to have: also send a ping and debounce when the app loads. This way we have faster feedback that the WS is connected

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
DutchBen: With the current solution the disconnect from being in suspense is solved.

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
DutchBen: Done.

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
```diff
@@ -23,11 +24,9 @@ const getWebSocket = async (url: string) => {
     }
 };
 
-// Recommended values are between 30 and 60 seconds
-// Note: When the browser tab is suspended, the browser will give this suspended tab approximately every 60 seconds time
-// to execute code. This means that the ping message will be sent every 60 seconds in suspended mode. This depends
-// on the browser and can differ.
-const PING_INTERVAL_MS = 45000;
+const PING_INTERVAL_MS = 30000;
+const NO_PONG_RECEIVED_INTERVAL_MS = 35000;
```
ricardovdheijden, line 28: Maybe `NO_PONG_RECEIVED_TIMEOUT`? (MS is always a given right?)

Could also introduce a `const SECONDS=1000` so you get `35 * SECONDS` as value

ricardovdheijden (approved):

Georgi2704 (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
DutchBen: Used the TIMEOUT suggestion. I do prefer MS in the name so didn't change that

PR #1380, opened 2024-08-29
Changes required for 1320 NMS breadcrumb with chassis name

Georgi2704:


DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoError/WfoError.tsx
```diff
@@ -27,7 +27,7 @@ export const WfoErrorWithMessage = ({ error }: WfoErrorWithMessageProps) => {
     const t = useTranslations('common');
     const { theme } = useOrchestratorTheme();
     const message =
-        error.length > 0
+        error && error.length > 0
```
ricardovdheijden, line 30: You can use the `?` to not repeat yourself here

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoError/WfoError.tsx
Georgi2704: TS is not happy

`TS18048: error.length is possibly undefined`

PR #1385, opened 2024-08-30
Generate library version when doing a build

Georgi2704:

PR #1394, opened 2024-09-03
1308 Adjust start task combobox size, so it doesn't hide list items

Georgi2704:


wouter1975 (approved):

packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartTaskComboBox.tsx
```diff
@@ -42,6 +42,7 @@ export const WfoStartTaskButtonComboBox = () => {
             options={comboBoxOptions}
             onOptionChange={handleOptionChange}
             isProcess={false}
+            width="600px"
```
ricardovdheijden, line 45: Consider using the `css` prop instead of a value. Therefore you need to add the className prop to WfoStartButtonComboBoxProps

packages/orchestrator-ui-components/src/configuration/version.ts
```diff
@@ -1 +1 @@
-export const ORCHESTRATOR_UI_LIBRARY_VERSION = '1.38.0';
+export const ORCHESTRATOR_UI_LIBRARY_VERSION = '1.38.1';
```
ricardovdheijden, line 1: i do not see a change in package.json, how come this one is bumped?

Georgi2704: When doing an npm release Version Packages will overwrite it to the newest one.

Georgi2704: Actually no.

My idea is that during build this file will be generated and then the deploy will continue with the generated file so it will display the new version. We can probably gitignore it as well.

DutchBen: It's from this PR. https://github.com/workfloworchestrator/orchestrator-ui-library/pull/1385

packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartTaskComboBox.tsx
DutchBen: Thought you meant use **!important** in css there for a while. Wasn't it **!discuss** instead of **!important**

ricardovdheijden: Yeah I used the should-never-use `!importtant` to mark something important for discussion. Might not be a good idea in a css context discussion here. Sticking to discuss from now on.

On topic: try to forward the css prop by either: 
- forwarding the `css` prop itself which has the type `SerializedStyles` from `@emotion/react`
- forwarding the `className` prop which also enables the css prop to be used

packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartButtonComboBox.tsx
```diff
@@ -48,12 +50,14 @@ export const WfoStartButtonComboBox = ({
             closePopover={() => setPopoverOpen(false)}
         >
             <EuiSelectable<StartComboBoxOption>
-                css={selectableStyle}
+                className={className}
+                css={[{ ...selectableStyle }]}
```
ricardovdheijden:
```suggestion
                css={selectableStyle}
```

ricardovdheijden (approved):

PR #1405, opened 2024-09-04
1381 Mouse over on cells to display full text

Georgi2704:


Georgi2704:
Can be closed after merging #1454

PR #1409, opened 2024-09-05
1408 set in sync should be authorized correctly

wouter1975:
Add authorization check to Set in sync action
Update UI Version text in mouse over
Added NPM generated version.ts file to .gitignore


DutchBen (approved):

DutchBen:
Add a changeset

PR #1425, opened 2024-09-12
Add sorting for subscription details workflows tab

Georgi2704:
image

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoRadioDropdown/WfoRadioDropdown.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoRadioDropdown/WfoRadioDropdown.tsx
@@ -8,13 +8,17 @@ import {
 } from '@elastic/eui';
 import { EuiRadioGroupOption } from '@elastic/eui/src/components/form/radio/radio_group';
 
-import { WfoRadioDropdownOption } from '@/types';
-
 export interface WfoRadioDropdownProps<T> {
     options: WfoRadioDropdownOption<T>[];
     selectedOption: WfoRadioDropdownOption<T>;
     onUpdateOption: (value: WfoRadioDropdownOption<T>) => void;
 }
+
+export type WfoRadioDropdownOption<T> = {
+    label: string;
+    id: string;
+    value: T;
+};
 
 export const WfoRadioDropdown = <T,>({
     options,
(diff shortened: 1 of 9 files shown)


packages/orchestrator-ui-components/src/components/WfoRadioDropdown/WfoRadioDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import { WfoRadioDropdownOption } from '@/types';
+
+export interface WfoRadioDropdownProps<T> {
+    options: WfoRadioDropdownOption<T>[];
+    selectedOption: WfoRadioDropdownOption<T>;
+    onUpdateOption: (value: WfoRadioDropdownOption<T>) => void;
+}
+
+export const WfoRadioDropdown = <T,>({
```
DutchBen, line 19: Is the , a typo?
```suggestion
export const WfoRadioDropdown = <T>({
```

packages/orchestrator-ui-components/src/components/WfoRadioDropdown/WfoRadioDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+    const [isOpen, setIsOpen] = useState(false);
+
+    const handleClose = () => {
+        if (isOpen) {
+            setIsOpen(false);
+        }
+    };
+    const handleSelectRadio = (id: string) => {
+        const updatedOption = options.find((option) => option.id === id);
+        setTimeout(handleClose, 100);
```
DutchBen, line 33: I would prefer a css transition to make this smoother if that is the intention

packages/orchestrator-ui-components/src/components/WfoRadioDropdown/WfoRadioDropdown.tsx
Georgi2704: Not a typo. Otherwise the compiler thinks it is a functional component <T>.

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoRadioDropdown/WfoRadioDropdown.tsx
Georgi2704: Elastic doesn't hide the Popover but removes it from the DOM, so I think It won't work with css transition.

packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -11,6 +11,12 @@ export type FieldValue = {
     value: string | number | boolean | null;
 };
 
+export type WfoRadioDropdownOption<T> = {
+    label: string;
+    id: string;
+    value: T;
+};
+
```
ricardovdheijden, line 19: I would suggest to keep this in the component, it is not domain data or something that is super-reusable (think about the SortOrder enum)

PR #1433, opened 2024-09-16
Added sorting for new tasks and workflow dropdowns

Georgi2704:
Screenshot 2024-09-16 at 15 57 35
Screenshot 2024-09-16 at 15 57 26

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartTaskComboBox.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartTaskComboBox.tsx
@@ -19,15 +19,13 @@ export const WfoStartTaskButtonComboBox = () => {
     const taskOptions = data?.startOptions || [];
 
     const comboBoxOptions: StartComboBoxOption[] = [...taskOptions]
-        .sort((taskA, taskB) =>
-            taskA.description.localeCompare(taskB.description),
-        )
         .map((option) => ({
             data: {
                 workflowName: option.name,
             },
-            label: option.description || '',
-        }));
+            label: option.description ?? option.name,
+        }))
+        .sort((taskA, taskB) => taskA.label.localeCompare(taskB.label));
 
     const handleOptionChange = async (selectedProduct: StartComboBoxOption) => {
         if (await isEngineRunningNow()) {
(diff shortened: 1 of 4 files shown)


wouter1975 (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/startOptions.ts
```diff
@@ -53,7 +53,9 @@ export type StartOptionsResponse<StartOption> = {
     startOptions: StartOption[];
 };
 
-type TaskOption = Pick<WorkflowDefinition, 'name' | 'description'>;
+type TaskOption = Pick<WorkflowDefinition, 'name'> & {
```
DutchBen, line 56: Isn't making the description non optional here a bit hacky? If we say a description may or may not be part of a workflow definition than it's hard to assume it will alway be present here right? Did you try making it non-optional on the WorkflowDefintion type to see what happens?

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartTaskComboBox.tsx
```diff
@@ -18,14 +18,16 @@ export const WfoStartTaskButtonComboBox = () => {
     const { data } = useGetTaskOptionsQuery();
     const taskOptions = data?.startOptions || [];
 
-    const comboBoxOptions: StartComboBoxOption[] = taskOptions.map(
-        (option) => ({
+    const comboBoxOptions: StartComboBoxOption[] = [...taskOptions]
+        .sort((taskA, taskB) =>
+            taskA.description.localeCompare(taskB.description),
+        )
+        .map((option) => ({
```
ricardovdheijden, line 25: Related to discussion below (@DutchBen ) -- maybe map first and then sort, just make sure the ones without description end up at the bottom (if they even exist)

packages/orchestrator-ui-components/src/rtk/endpoints/startOptions.ts
ricardovdheijden: Agree, `discuss`

Issue #1363, opened 2024-08-22
Keep main menu items highlighted for workflows/tasks/subscriptions and add in H1 header


wouter1975:
Use similar navigation and header style as used in NMS GUI.
Make the header consistent throughout the whole app (LIR/opensource/service ticket/NMS)

See https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1316

PR #1435, opened 2024-09-17
1363 Adds WfoMultilineCell and WfoTruncateCell

ricardovdheijden:
#1363 

Adds:
WfoMultilineCell
WfoTruncateCell

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoMultilineCell.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoMultilineCell.tsx
@@ -1,0 +1,13 @@
+import React, { FC, ReactElement } from 'react';
+
+export type WfoMultilineCellProps = {
+    children?: ReactElement | string | null;
+};
+
+export const WfoMultilineCell: FC<WfoMultilineCellProps> = ({ children }) => {
+    if (!children) {
+        return null;
+    }
+
+    return <p css={{ textWrap: 'wrap' }}>{children}</p>;
+};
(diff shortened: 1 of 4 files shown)


packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoMultilineCell.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { FC, ReactElement } from 'react';
+
+export type WfoMultilineCellProps = {
+    children?: ReactElement | string | null;
```
DutchBen, line 4: Is use the ReactNode for children, that includes string, number etc.. What is the benefit of ReactElement here?

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoMultilineCell.tsx
ricardovdheijden: ReactNode  is throwing compile errors that it is not allowed in an FC

wouter1975 (approved):

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoMultilineCell.tsx
mrijk: Looks a bit weird that children is plural, but we got an Element or a string or null. I was expecting a list.

PR #1438, opened 2024-09-17
1376 inline note field

DutchBen:


packages/orchestrator-ui-components/src/components/WfoInlineNoteEdit/WfoInlineNoteEdit.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const triggerNoteModifyWorkflow = () => {
+        const noteModifyPayload = [
+            { subscription_id: subscriptionId },
+            { note },
+        ];
+        startProcess({
+            workflowName: 'modify_note',
+            userInputs: noteModifyPayload,
+        });
+    };
```
ricardovdheijden, line 28: Move this block under the `useState`

DutchBen: Please use the code suggestion option for this preference related suggestions.

packages/orchestrator-ui-components/src/components/WfoInlineNoteEdit/WfoInlineNoteEdit.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    return (
+        <EuiInlineEditText
+            inputAriaLabel="Edit note"
+            value={note}
+            onChange={(e: ChangeEvent<HTMLInputElement>) => {
+                setNote(e.target.value);
+            }}
+            onCancel={() => setNote(initialNote)}
+            onSave={() => triggerNoteModifyWorkflow()}
```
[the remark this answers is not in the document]
DutchBen: The note is a value we get from state so imo it doesn't make the function clearer to pass it in. The value isn't passed into the save function.

wouter1975 (approved):

PR #1454, opened 2024-09-23
Render mouseover tooltip for the new tables

Georgi2704:


packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDataCell.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { ReactNode } from 'react';
+
+import { EuiToolTip } from '@elastic/eui';
+
+interface WfoDataCellProps {
+    showTooltip: boolean;
+    content: ReactNode;
```
ricardovdheijden, line 7: If i understand this correct this is the content inside the tooltip ietself and not the actual data in the cell?
Therefore I would suggeste to rename this into `tooltipContent`

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoBadges/WfoBadge/WfoBadge.tsx
```diff
@@ -18,7 +18,7 @@ export const WfoBadge: FC<WfoBadgeProps> = ({
     ...restProps
```
ricardovdheijden, line 18: don't forget the changeset

packages/orchestrator-ui-components/src/components/WfoBadges/WfoBadge/WfoBadge.tsx
```diff
@@ -18,7 +18,7 @@ export const WfoBadge: FC<WfoBadgeProps> = ({
...
 }) => (
     <EuiBadge {...restProps}>
-        <EuiText color={textColor} size={size ?? 'xs'}>
+        <EuiText title="" color={textColor} size={size ?? 'xs'}>
```
DutchBen, line 21: What does the empty title prop? 
```suggestion
        <EuiText color={textColor} size={size ?? 'xs'}>
```

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDataCell.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { EuiToolTip } from '@elastic/eui';
+
+import { useOrchestratorTheme } from '@/hooks';
+
+interface WfoDataCellProps {
+    tooltipContent: ReactNode;
+    children: ReactNode;
+}
+
+export const WfoDataCell = ({ tooltipContent, children }: WfoDataCellProps) => {
```
DutchBen, line 12: I think we said - againt my preference :) - we'd use FC here

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoBadges/WfoBadge/WfoBadge.tsx
Georgi2704: It prevents showing the default browser tooltip. We don't want those showing.

packages/orchestrator-ui-components/src/icons/WfoCheckmarkCircleFill.tsx
```diff
@@ -14,7 +14,6 @@ export const WfoCheckmarkCircleFill: FC<WfoIconProps> = ({
         version="1.1"
         xmlns="http://www.w3.org/2000/svg"
     >
-        <title>icon/checkmark-circle-fill</title>
```
ricardovdheijden, line 17: Adjusting SVG files might not be the best solution. Can't you apply a css rule?

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/icons/WfoCheckmarkCircleFill.tsx
Georgi2704: I don't think it is possible only with CSS. I've tried before but didn't succeed. 

And if there is a way with css, we always have to hide the default tooltip (title). Isn't it better to just remove it from the svg ?

Issue #459, opened 2023-11-15
Single click sorting on table


wouter1975:
If no filtering is allowed on a table column, the filter modal should not be shown to the user, but the sorting is immediately executed. 

Cycle through the following actions

1. sort ascending --> update icon in header
2. sort descending --> update icon in header

- [x] Uncouple the sorting and filtering action. Thus no sorting from the modal.

Icons for sorting arrows, see below. Note that not-sortable column will show on hoover a forbidden icon.

![Image](https://github.com/user-attachments/assets/7fa5d5b2-6d6a-4126-ae50-c1181d5e8670)

wouter1975:
won't fix for now

Issue #315, opened 2023-09-28
Sort icon not visible if column header is ~ same as columnwidth


wouter1975:
Seen on process list page, the sorting arrow is (partly) lost in the view where the column header roughly the same width as the column itself. 

![Image](https://github.com/workfloworchestrator/orchestrator-ui/assets/83226590/9f71dcaf-c3ac-43d9-8133-53ab7545c0dd)

Solution directions?
* Either add a minimal width
* Or crop the header text such that the sorting arrow can always be shown
- [ ] solution proposed by Ruben, is to overlay the sorting arrow on top of the column header

Georgi2704:
@wouter1975 I believe this has been already fixed:
![Image](https://github.com/user-attachments/assets/29dd70f2-692d-4359-af7c-43a9c2b7fe6b)

PR #1460, opened 2024-09-26
459 single click sorting + 315 sort icon always visible

ricardovdheijden:
#459
#315 

- Implements single click sorting: for sortable columns the sortable icon only is visible on hover or focus
- Adds heroicons subfolder in icons and a wrapper with margins to match the other icons in the project

- Makes sort icon always visible

image


packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderCell/WfoSortDirectionIcon.tsx
```diff
@@ -1,18 +1,30 @@
 import React, { FC } from 'react';
 
 import { useOrchestratorTheme } from '@/hooks';
-import { WfoArrowNarrowDown, WfoArrowNarrowUp } from '@/icons';
+import { WfoArrowNarrowDown, WfoArrowNarrowUp, WfoArrowsUpDown } from '@/icons';
 import { SortOrder } from '@/types';
 
+export const SORTABLE_ICON_CLASS = 'sortableIcon';
```
Georgi2704, line 7: I would place this in styles.ts.

Georgi2704 (approved):

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderCell/WfoSortDirectionIcon.tsx
ricardovdheijden: Yes, that is better indeed

Georgi2704 (approved):

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderCell/styles.ts
```diff
@@ -1,6 +1,11 @@
 import { css } from '@emotion/react';
 
 import { WfoTheme } from '@/hooks';
+import { SortOrder } from '@/types';
+
+export const HEADER_CELL_TITLE_BUTTON = 'headerCellTitleButton';
```
DutchBen, line 6: Are these default eui classes? Maybe postfix the const name with _CLASS

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/utils.spec.ts
```diff
@@ -21,59 +25,83 @@ const tableColumnConfig: WfoTableColumnConfig<TestObject> = {
...
+            if (
+                result.name?.columnType === ColumnType.DATA &&
+                result.age?.columnType === ColumnType.DATA
+            ) {
+                expect(result.name.isSortable).toEqual(true);
+                expect(result.name.isFilterable).toEqual(true);
+                expect(result.age.isSortable).toEqual(true);
+                expect(result.age.isFilterable).toEqual(true);
+            } else {
+                // Preventing silently skipping above expects
```
DutchBen, line 75: In what case would you end up in this branch? Don't you always get the same results? I see this was already there but maybe we can remove it

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/utils.spec.ts
ricardovdheijden: TypeScript is unable to determine the column type of the `name` and `age` columns, the check is needed to avoid TSC errors. We need the else branch to make the unit test fail when skipped

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderCell/styles.ts
ricardovdheijden: done

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/utils.spec.ts
```diff
@@ -21,59 +25,83 @@ const tableColumnConfig: WfoTableColumnConfig<TestObject> = {
...
-            expect(result.name.isSortable).toEqual(false);
-            expect(result.name.isFilterable).toEqual(false);
-            expect(result.age.isSortable).toEqual(false);
-            expect(result.age.isFilterable).toEqual(false);
-        } else {
-            // Preventing silently skipping above expects
-            throw Error('Some of the fields are not data fields');
-        }
+            // Then
+            if (
```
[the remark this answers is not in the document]
ricardovdheijden: TypeScript is unable to determine the column type of the name and age columns when it comes back from the mapper function, the check is needed to avoid TSC errors. We need the else branch to make the unit test fail when skipped.

Issue #735, opened 2024-01-31
[Low] clicking the x icon in search bar is not resetting the data in the processes table


wouter1975:
clicking the x icon in search bar is not resetting the data in the processes table, works for subscriptions table

with a column search active: https://orchestrator-v2.dev.automation.surf.net/workflows?pageSize=10&pageIndex=3&sortBy=field-productTag_order-DESC&queryString=lastStatus%3Afailed

tjeerddie:
this seems to be fixed already, you do need to remove focus from the search box

PR #1497, opened 2024-10-14
735 Triggers onUpdateQueryString when search field is erased

ricardovdheijden:
#735 

- Triggers onUpdateQueryString when search field is erased.
- Adds queryIsValid prop (optional prop) -- it was hardcoded to true

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoSearchBar/WfoSearchField.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSearchBar/WfoSearchField.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import { useTranslations } from 'next-intl';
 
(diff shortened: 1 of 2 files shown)


packages/orchestrator-ui-components/src/components/WfoSearchBar/WfoSearchField.tsx
```diff
@@ -22,15 +24,20 @@ export const WfoSearchField = ({
         getFormFieldsBaseStyle,
     );
 
-    const queryIsValid = true; // Query validation turned of for now until ESQueries can be sent to the backend
+    const handleSearch = useRef((queryText: string) =>
```
DutchBen, line 27: FMI. What's the benefit of using useRef here instead of useCallback?

ricardovdheijden: The reference keeps changing and triggers a renderloop. useCallback also requires a dependency array (just like useEffect)

wouter1975 (approved):

PR #1499, opened 2024-10-15
Adjusted FieldValue type to include ReactNode for NMS

Georgi2704:


DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoSubscription/utils/utils.ts
```diff
@@ -25,7 +25,15 @@ export const getFieldFromProductBlockInstanceValues = (
     const nameValue = instanceValues.find(
         (instanceValue) => instanceValue.field === field,
     );
-    return nameValue?.value ?? '';
+    const value = nameValue?.value;
```
DutchBen, line 28: So the value according to the type can be `ReactNode | (string | number | boolean | null)` but you only return it here if it's one of `string | number | boolean`. That doesn't seem right. What happens if you happen to use this utility function when you do expect a ReactNode or null? 
I know this was already there but the name of the function is getFieldFromProductBlockInstanceValues. That doesn't cover it. Maybe rename the function or make a new one that does what you want?

Georgi2704: If I have to make a new one it will be a lot of unnecessary duplicated code for one type. The component WfoProductBlockKeyValueRow already suggests that ReactNode should be valid. There are also type checks in it. 

`
    const WfoProductBlockValue: FC<{ value: FieldValue['value'] }> = ({
        value,
    }) => {
        if (typeof value === 'boolean') {
            return <EuiBadge>{value.toString()}</EuiBadge>;
        } else if (Array.isArray(value)) {
            const result = value.join(', ');
            return <>{result}</>;
        } else {
            return <>{value}</>;
        }
    };

`

Will `ReactNode | (string | number | boolean)` | null or  `ReactNode | string | number | boolean | null `
be a valid type ?

wouter1975 (approved):

tjeerddie (approved):

Issue #1500, opened 2024-10-16
NPM Major version upgrades


DutchBen:
These npm packages have major version upgrades that we might want to implement:
```
Package                           Current  Wanted         Latest  
@turbo/gen                         1.13.4  1.13.4          2.1.3  --> step 1
@typescript-eslint/eslint-plugin   6.21.0  6.21.0          8.9.0 --> step 2
eslint                             8.56.0  8.57.1         9.12.0  --> step 2
graphql-request                     6.1.0   6.1.0          7.1.0 --> step 3
react-redux                         8.1.3   8.1.3          9.1.2 ---> wait, yolo --> step 4
turbo                              1.13.4  1.13.4          2.1.3 --> step 1
```

Notes:
- graphql-request When previously trying to upgrade graphql-request the build command stopped working demanding the `moduleresolution` to be changed to `bundler` leading to other changes to the tsconfig setup https://www.typescriptlang.org/tsconfig/#moduleResolution
- react-redux: When previously trying to upgrade react-redux it lead to errors about not being able to dynamically load react. This probably related to the way we have a lot of react logic in our lib
- eslint: Eslint > 9 configuration files require a different flat format 
- turbo/turbo-gen: Haven't tried upgrading these yet

In addition to these major upgrades these deprecation messages are currently showing up when runnin 'npm i'. Check them after these major version upgrades:
```
npm warn deprecated abab@2.0.6: Use your platform's native atob() and btoa() methods instead
npm warn deprecated domexception@4.0.0: Use your platform's native DOMException instead
npm warn deprecated core-js@2.6.12: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.```

PR #1501, opened 2024-10-16
1487 updates all minor npm package versions and some major

DutchBen:
Updates all minor version upgrades and some major. ElasticEUI updated to latest major version. This finally supports TS 5 so an override is removed. 

After this update a there are 2 `npm audit` warnings left
- cookie < v0.7 used in next-auth: this will be fixed upstream after this pr: https://github.com/nextauthjs/next-auth/pull/12046
- micromatch: this is part of the lodash version used in uniforms. Wont fix sinc

Some packages have major version upgrades that are moved to another ticket: https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1500

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     ],
     "dependencies": {
         "@elastic/eui": "^97.0.0",
-        "@storybook/testing-library": "^0.2.2",
         "react-redux": "^8.1.3"
     }
 }
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     ],
     "dependencies": {
         "@elastic/eui": "^97.0.0",
+        "@testing-library/react": "^16.0.1",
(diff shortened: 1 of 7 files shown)


Georgi2704 (approved):

package.json
```diff
@@ -24,5 +24,10 @@
     "workspaces": [
         "apps/*",
         "packages/*"
-    ]
+    ],
+    "dependencies": {
+        "@elastic/eui": "^97.0.0",
+        "@testing-library/react": "^16.0.1",
+        "react-redux": "^8.1.3"
```
ricardovdheijden, line 31: As far as i know these do not belong here, since this is the package.json of the mono repo itself. Such dependencies would be more an app related dependency. `Discussion`

PR #1507, opened 2024-10-17
Fix: Add onchange on the useEffect state in order to set the default  #2110

Igoranze:
Add onchange on the useEffect state in order to set the default value to the userinput


wouter1975 (approved):

wouter1975 (approved):

.changeset/many-dogs-dream.md
```diff
@@ -0,0 +1,5 @@
+---
```
ricardovdheijden, line 1: Looks like a duplicate changeset, you can remove either of the 2 files

ricardovdheijden (approved):

PR #1516, opened 2024-10-22
1404 Changes to WfoFilterTabs for NMS chassis detail tabs

Georgi2704:


.changeset/gentle-wombats-float.md
```diff
@@ -0,0 +1,5 @@
+---
+'@orchestrator-ui/orchestrator-ui-components': patch
+---
+
+1404 Changes to WfoFilterTabs for NMS chassis detail tabs
```
DutchBen, line 5: This message should not reference an internal SURF implementation
```suggestion
1404 Allows WfoFilterTabs to accept append property that is passed on to EuiTab component
```

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoFilterTabs/styles.ts
```diff
@@ -0,0 +1,13 @@
+import { css } from '@emotion/react';
+
+import { WfoTheme } from '@/hooks';
+
+export const getStyles = ({ theme }: WfoTheme) => {
+    const tabStyle = css({
+        '.euiTab__append': {
+            marginLeft: `-${theme.size.xs}`,
```
ricardovdheijden, line 8: I think we should avoid negative margins and paddings and such as much as possible. `discussion`

ricardovdheijden (approved):

PR #1517, opened 2024-10-22
Fix: Empty value dropdown override

Igoranze:
Bug introduces when fixing the onchange on dropdown AGGSP, this fixes the bug


packages/orchestrator-ui-components/src/components/WfoForms/formFields/SelectField/SelectField.tsx
```diff
@@ -98,7 +98,9 @@ export function UnconnectedSelectField({
         useWithOrchestratorTheme(getSelectFieldStyles);
 
     useEffect(() => {
-        onChange(selectedValue?.value);
+        if (selectedValue && selectedValue.value !== 'undefined') {
+            onChange(selectedValue?.value);
+        }
     }, []);
```
DutchBen, line 104: Fix the linting error
```suggestion
    }, [onChange, selectedValue]);
```

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/SelectField/SelectField.tsx
ricardovdheijden: Do verify if the app still works when adding this. It might lead to a render loop. I tried to fix similar cases by applying a useRef of those dependencies that causes the render loop

packages/orchestrator-ui-components/src/components/WfoForms/formFields/SelectField/SelectField.tsx
```diff
@@ -98,7 +98,9 @@ export function UnconnectedSelectField({
         useWithOrchestratorTheme(getSelectFieldStyles);
 
     useEffect(() => {
-        onChange(selectedValue?.value);
+        if (selectedValue && selectedValue.value !== 'undefined') {
```
ricardovdheijden, line 101: `selectedValue && selectedValue.value` can be replaced with `selectedValue?.value`. And the question mark in the onChange can be removed

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/SelectField/SelectField.tsx
Igoranze: Thanks, but if I do this I get errors on runtime, not sure how to fix it properly in the FE

florisie (approved):

Issue #1402, opened 2024-09-03
[NMS] Add action Set Corelink in maintenance


wouter1975:
Would be good to use the existing modify workflow to toggle the maintenance mode of a corelink as a lightweight action from NMS GUI.

The payload of the workflow shall be:
https://orchestrator.dev.automation.surf.net/api/processes/modify_core_link_maintenance
[{"subscription_id":"95ce5b7f-e66b-4926-801e-7a7c1ba79a52"},{"maintenance_mode":true}]

or
[{"subscription_id":"95ce5b7f-e66b-4926-801e-7a7c1ba79a52"},{"maintenance_mode":false}]

* Thus conditional menu item is shown if subscription type is `corelink`
* the active state of the core link is reflected in the submenu, showing the option being ticked.
* Red Icon is should if maintenance_mode = true (NWD is using this icon)

Would the following design be possible to implement: UPDATE: use a modal with a header `Change maintenance mode` with a button which either has the text `Enable maintenance mode` or `Disable maintenance mode`

![Image](https://github.com/user-attachments/assets/e204a349-c841-47c4-b4a8-eee3d38d881f)

- [x] Add isAllowed --> should disable the menu item when not allowed
  - [x] Using existing resource: '/orchestrator/subscriptions/modify/[subscriptionId]',
- [x] When maintenance mode is not available: show "no actions available"

PR #1520, opened 2024-10-29
1402 table row context menu

ricardovdheijden:
#1402 

Adds context menu to be used in the table. It exposes a callback that is triggered when the menu opens (for optionally triggering a REST call to populate the menu)

Example usage:
image


Igoranze (approved):
LGTM

Georgi2704 (approved):

PR #1527, opened 2024-10-30
1500 npm major versions

DutchBen:
Updates some major versions including turbo 2.* 

- I have left the upgrade to next 15 for later since it also comes with an update to React 19.
- I have left the upgrade of eslint to version 9 till we update to next 15 that supports it
- Upgrading to the latest graphql-request would force use to change the moduleResolution option to 'bundler' and also force our user to do so I'm moving that to a different ticket
- After this merge we have 2 audit warnins left
 -- Lodash.template: comes from uniforms which we are nog going to fix for now. 
-- elliptic: used upstream by storybook. Waiting for them to fix it.

Screenshot 2024-10-30 at 16 18 12


ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/deprecated/ImsNodeIdField.tsx
```diff
@@ -80,8 +80,8 @@ function ImsNodeId({
         isLoading && locationCode
             ? t('widgets.node_select.nodes_loading')
             : nodes.length
-              ? t('widgets.node_select.select_node')
-              : t('forms.widgets.node_select.no_nodes_placeholder');
+            ? t('widgets.node_select.select_node')
+            : t('forms.widgets.node_select.no_nodes_placeholder');
```
ricardovdheijden, line 84: This becomes very unreadable, can you try to fix the rule that is responsible for this?

ricardovdheijden (approved):

.changeset/healthy-days-shout.md
```diff
@@ -0,0 +1,7 @@
+---
+'@orchestrator-ui/orchestrator-ui-components': patch
```
Georgi2704, line 2: isn't that more of a minor upgrade than a patch ?

wouter1975 (approved):

Georgi2704 (approved):

PR #1533, opened 2024-11-04
Make last child of product block key value item without bottom border

Georgi2704:
Before:

Screenshot 2024-10-31 at 15 21 06
Screenshot 2024-10-31 at 15 23 10

After:
Screenshot 2024-10-31 at 15 20 47
Screenshot 2024-10-31 at 15 22 47


packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock/styles.ts
```diff
@@ -43,6 +43,9 @@ export const getStyles = (wfoTheme: WfoTheme) => {
         '&:first-child': {
             borderTop: `solid 1px ${theme.colors.lightShade}`,
         },
+        '&:last-child': {
+            borderBottom: `solid 0px ${theme.colors.lightShade}`,
```
DutchBen, line 47: I think you can do 'none' here instead

ricardovdheijden: +1

Mark90 (approved):

PR #1550, opened 2024-11-11
1506 alex toggle

DutchBen:


packages/orchestrator-ui-components/src/components/WfoKeyValueTable/styles.ts
```diff
@@ -44,6 +44,8 @@ export const getStyles = ({ theme }: WfoTheme) => {
     const valueColumnStyle = css({
         borderTopRightRadius: radius,
         borderBottomRightRadius: radius,
+        paddingTop: theme.base / 2,
+        paddingBottom: theme.base / 2,
```
ricardovdheijden, line 48: I think it might be better to use `theme.size.s` ( which is also 8px ), because `/2` could result in half pixels if `theme.base` is ever changed in the future.

packages/orchestrator-ui-components/src/components/WfoTableCodeBlock/styles.ts
```diff
@@ -0,0 +1,13 @@
+import { css } from '@emotion/react';
+
+export const getStyles = () => {
+    const tableCodeBlockMarginStyle = css({
+        margin: 15,
+        marginTop: 20,
+        marginRight: 20,
```
ricardovdheijden, line 7: I think it might be better to use values from the theme here, else introduce constants. `discussion`

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoKeyValueTable/styles.ts
DutchBen: done

packages/orchestrator-ui-components/src/components/WfoTableCodeBlock/styles.ts
DutchBen: Done

PR #1562, opened 2024-11-13
1553 process limit

DutchBen:


Georgi2704 (approved):

wouter1975 (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionDetail.ts
```diff
@@ -50,7 +51,7 @@ fragment SubscriptionDetail on SubscriptionInterface {
         subscriptionInstanceId
         inUseByRelations
     }
-    processes(sortBy: { field: "startedAt", order: ASC }) {
+    processes(first: ${NUMBER_OF_ITEMS_REPRESENTING_ALL_ITEMS}, after: 0, sortBy: { field: "startedAt", order: ASC }) {
```
ricardovdheijden, line 54: Do we always need everything at once? Because otherwise using a graphql variable would be preferred.

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/subscriptionDetail.ts
DutchBen: Yes. There is no pagination so we need it all

Issue #1577, opened 2024-11-18
Input form checkbox has a click-offset to the right, not on the icon but on text


wouter1975:
One needs to click just outside the checkbox to enable/disable it 
![Image](https://github.com/user-attachments/assets/e1823ae7-78db-46af-ac5b-4f1bebb61ffa)

PR #1589, opened 2024-11-22
1577 Fixes uncheckable checkbox (only the label was clickable)

ricardovdheijden:
#1577 

Fixes uncheckable checkbox (only the label was clickable)

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoForms/formFields/BoolFieldStyling.ts
+++ b/packages/orchestrator-ui-components/src/components/WfoForms/formFields/BoolFieldStyling.ts
@@ -12,10 +12,6 @@ export const boolFieldStyling = css`
       z-index: unset;
     }
 
-    input[type="checkbox"] {
-      display: none;
-    }
-
     label {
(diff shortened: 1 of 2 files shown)


Georgi2704 (approved):

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoForms/formFields/BoolFieldStyling.ts
```diff
@@ -12,10 +12,6 @@ export const boolFieldStyling = css`
       z-index: unset;
     }
 
-    input[type="checkbox"] {
```
DutchBen, line 15: This looks like it's put there intentionally so I am a bit worried we are going to introduce another problem. Any guesses to why it was put here in the first place?

ricardovdheijden: I share the worries, but according to the git history it seems introduced in the v1 implementation.

PR #1601, opened 2024-11-27
1544 resize columns

DutchBen:
First part of the resizable columns. Still missing:
- Storing resized columns in local storage so they presist

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDragHandler.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDragHandler.tsx
@@ -8,6 +8,8 @@ import { useWithOrchestratorTheme } from '@/hooks';
 import type { onUpdateColumWidth } from './WfoTableHeaderRow';
 import { getWfoTableStyles } from './styles';
 
+const MINIMUM_COLUMN_WIDTH = 50;
+
 interface WfoDragHandlerProps {
     headerRowRef: React.RefObject<HTMLTableRowElement>;
     fieldName: string;
@@ -49,13 +51,15 @@ export const WfoDragHandler: FC<WfoDragHandlerProps> = ({
 
                         onUpdateColumWidth(
                             fieldName,
-                            newWidth > 50 ? newWidth : 50,
+                            newWidth > MINIMUM_COLUMN_WIDTH
+                                ? newWidth
+                                : MINIMUM_COLUMN_WIDTH,
                         );
                         resetPosition();
                     }
                 }}
             >
-                <div css={dragAndDropStyle}>&nbsp;</div>
+                <div css={dragAndDropStyle} />
             </Draggable>
         </div>
     );
--- a/packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/styles.ts
+++ b/packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/styles.ts
@@ -140,6 +140,9 @@ export const getWfoTableStyles = ({ theme, isDarkThemeActive }: WfoTheme) => {
                 ? theme.colors.mediumShade
                 : theme.colors.header,
         },
+        '&::after': {
+            content: '""',
+        },
     });
     return {
         tableContainerStyle,


packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderRow.tsx
```diff
@@ -83,36 +95,73 @@ export const WfoTableHeaderRow = <T extends object>({
...
-                            <div>{columnConfig.label?.toString() || ''}</div>
+                                                onUpdateColumWidth(
+                                                    fieldName,
+                                                    newWidth,
+                                                );
+                                            }
+                                        }}
+                                    >
+                                        &nbsp;
+                                    </div>
```
ricardovdheijden, line 141: Was it not possible to use the boundary between 2 cells as point to drag on? I think that is why it becomes impossible to resize very narrow colums: colums where the header text cannot fit in

packages/orchestrator-ui-components/package.json
```diff
@@ -52,6 +52,7 @@
         "object-hash": "^3.0.0",
         "prism-themes": "^1.9.0",
         "react-diff-view": "^3.2.0",
+        "react-draggable": "^4.4.6",
```
ricardovdheijden, line 55: Why not using the EUI built-in way of resizing instead of adding a new dependency?
https://eui.elastic.co/#/layout/resizable-container#custom-resize-logic
That one is also more a11y-compliant

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDragHandler.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useState } from 'react';
```
ricardovdheijden, line 1: Before merging: verify if the sort buttons are still working, it didn't work on the preview link (but that could be a data issue)

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderRow.tsx
DutchBen: I have since fixed the resizing of smaller columns and put in a minimum width. There is a draghandler component now that makes it possible/easier to show a movable line

packages/orchestrator-ui-components/package.json
DutchBen: That component doesn't have a standard callback that we need to use to update the column widths after we are done dragging. Also we have moved the table outside of EUI so it doesn't feel logical to move components back in

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderRow.tsx
```diff
@@ -83,36 +95,73 @@ export const WfoTableHeaderRow = <T extends object>({
...
                                             : undefined
                                     }
                                 >
                                     {columnConfig.label?.toString()}
                                 </WfoTableHeaderCell>
-                            </th>
-                        );
-                    }
+                                {onUpdateColumWidth && (
+                                    <div
```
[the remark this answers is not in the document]
DutchBen: Done

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderCell/styles.ts
```diff
@@ -81,11 +81,7 @@ export const getWfoBasicTableStyles = ({ theme }: WfoTheme) => {
         display: 'flex',
         justifyContent: 'flex-start',
         alignItems: 'center',
-
-        [`.${HEADER_CELL_TITLE_BUTTON_CLASS}:has(+ .${HEADER_CELL_SORT_BUTTON_CLASS}:focus-visible)`]:
-            {
-                overflow: 'hidden',
-            },
+        overflow: 'hidden',
```
ricardovdheijden, line 84: This probably breaks the functionality where the sort button should be visible when colums are less wide than the header text uses

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDragHandler.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+                        onUpdateColumWidth(
+                            fieldName,
+                            newWidth > 50 ? newWidth : 50,
+                        );
+                        resetPosition();
+                    }
+                }}
+            >
+                <div css={dragAndDropStyle}>&nbsp;</div>
```
ricardovdheijden, line 58: As far as I know you can add this space to the content prop in css or something with the `after` pseudo class

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDragHandler.tsx
```diff
@@ -0,0 +1,62 @@
...
+                        const thElement = headerRowRef.current.querySelector(
+                            `th[data-field-name="${fieldName}"]`,
+                        ) as HTMLTableCellElement;
+                        const startWidth =
+                            thElement.getBoundingClientRect().width;
+                        const newWidth = startWidth + data.x;
+
+                        onUpdateColumWidth(
+                            fieldName,
+                            newWidth > 50 ? newWidth : 50,
```
Georgi2704, line 52: const MINIMUM_COLUMN_WIDTH = 50 or something similar

Georgi2704 (approved):

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTableHeaderCell/styles.ts
DutchBen: Checked. Still seems to work

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDragHandler.tsx
DutchBen: Done

packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoDragHandler.tsx
DutchBen: done

Issue #692, opened 2024-01-24
Add sentry reporting


wouter1975:
In SURF specifiek deployment add sentry reporting, with possibly playback option (see NWD).

- [x] add in surf app
- [x] Add SurfErrorMonitoringProvider to app (implementing Sentry)
~- [ ] Test in Dev (sourcemaps)~ -> sourcemaps currently not working
- [x] Add WfoErrorMonitoringProvider, integrate with SurfErrorMonitoringProvider to library and add some errors to be logged
  - [x] Websockets
  - [x] Forms (cancel/next button issue)
- [ ] Remove test-page before closing story
- [x] Determine Sentry Sample rates (see comments in MR 275 in Gitlab)
- [x] Determine values to be configurable (see comments in MR 275 in Gitlab)
- [x] Update Okteto config for new env vars

Release to **Production**:
- [ ] Add secrets:
  - [ ] `SENTRY_AUTH_TOKEN` (does not exist in code, but it used by Sentry plugin)
- [ ] Add env vars:
  - [ ] `SENTRY_ENVIRONMENT`
  - [ ] `SENTRY_DSN_URL`
  - [ ] `TRACING_ENABLED` 
  - [ ] `SENTRY_ORGANIZATION`
  - [ ] `SENTRY_PROJECT`
  - [ ] `SENTRY_URL`
  - [ ] `TRACE_SAMPLE_RATE`

PR #1611, opened 2024-12-02
692 Error monitoring

ricardovdheijden:
#692 

Adds a `WfoErrorMonitoringProvider` to be implemented in the wfo-ui app in `_app.tsx`.
It is providing a handler to report errors to remote logging / tracing (Sentry for example)

The `useWfoErrorMonitoring()` hook exposes 2 functions: `reportError` and `reportMessage`. The provider is introduced to not tightly couple to Sentry. Implementing the monitoring is optional. The `WfoErrorMonitoringProvider` does not need to be added to `_app.tsx`.

### DIFF
--- a/packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
+++ b/packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
@@ -82,7 +82,7 @@ const streamMessagesApi = orchestratorApi.injectEndpoints({
                             orchestratorApi.util.invalidateTags([cacheTag]);
                         dispatch(cacheInvalidationAction);
                     } else {
-                        throw new Error(
+                        console.error(
                             `Trying to invalidate a cache entry with an unknown tag: ${cacheTag.type}`,
                         );
                     }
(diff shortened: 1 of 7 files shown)


packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
```diff
@@ -97,9 +97,12 @@ const streamMessagesApi = orchestratorApi.injectEndpoints({
                 const getDebounce = (delay: number) => {
                     return debounce(() => {
                         webSocket.close();
-                        // note: websocket.close doesnt trigger the onclose handler when losing
+                        // note: websocket.close doesn't trigger the onClose handler when losing
```
DutchBen, line 100:
```suggestion
                        // note: websocket.close doesn't trigger the onClose handler when closing
```

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
```diff
@@ -97,9 +97,12 @@ const streamMessagesApi = orchestratorApi.injectEndpoints({
...
                         // internet connection so we call the cleanup event from here to be sure it's called
                         cleanUp();
+                        throw new Error(
```
DutchBen, line 103: Doesn't this mean that when no wfoErrorMonitoring is implemented that these error become uncaught and will crash the application?

DutchBen (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/streamMessages.ts
ricardovdheijden: Good point, will remove the error-throwing. Implementing it later

PR #1640, opened 2024-12-18
Allow Create LIR prefixes workflow to start with pre-filled variables

Georgi2704:
Screenshot 2024-12-18 at 20 15 47
Screenshot 2024-12-18 at 20 16 09
Screenshot 2024-12-18 at 20 16 26

### DIFF
--- a/packages/orchestrator-ui-components/src/pages/processes/WfoStartProcessPage.tsx
+++ b/packages/orchestrator-ui-components/src/pages/processes/WfoStartProcessPage.tsx
@@ -75,22 +75,20 @@ const getInitialProcessPayload = ({
     prefixlen,
     prefix_min,
 }: StartProcessPageQuery): StartWorkflowPayload | undefined => {
-    if (productId) {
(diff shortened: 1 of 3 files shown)


DutchBen:
This works but I'm not sure it's fixed at the right place. On the other hand. We can live with it for now I guess

packages/orchestrator-ui-components/src/pages/processes/WfoStartProcessPage.tsx
```diff
@@ -75,6 +85,14 @@ const getInitialProcessPayload = ({
             subscription_id: subscriptionId,
         };
     }
+    if (productId && prefix && prefixlen && prefix_min) {
```
ricardovdheijden, line 88: Isn't this unreachable code? the if on line 78 above `if (productId) {` would trigger true before this if is reached

DutchBen (approved):

wouter1975 (approved):

Issue #1479, opened 2024-10-08
Add export button to the chassis / services table


wouter1975:
- [x] Export customers tab
- [x] Export services tab
- [x] Export port list tab (only the collapsed rows)

- [x] use the expand/collapse action style for the export button, just above the table

PR #1649, opened 2025-01-03
1479 grouped table and csv download updates

ricardovdheijden:
#1479 

Changes to support functionality in the SURF specific app but also useful for consumers of the library:
- GroupedTable: The table has a small header with a button "Collapse / Expand", this can now be overridden
- csvDownload: Changed the parameters of initiateCsvFileDownload for directly exporting data to CSV
- Introduced a utility type to help ensuring that all the props of an object will be of type "string". Typically useful for the csvDownload functionality

### DIFF
--- a/.changeset/real-planets-bake.md
+++ b/.changeset/real-planets-bake.md
@@ -1,0 +1,5 @@
+---
+'@orchestrator-ui/orchestrator-ui-components': patch
+---
+
+1479 GroupedTable: The table has a small header with a button "Collapse / Expand", this can now be overridden. csvDownload: the function "csvDownloadHandler" is async and allows to fetch data for the consumer before a CSV download is started by the browser. The newly added "csvDownloadHandlerSync" is introduced if the requested data is already present. Introduced a utility type to help ensuring that all the props of an object will be of type "string". Typically useful for the csvDownload functionality
--- a/.changeset/real-planets-bake.md
+++ b/.changeset/real-planets-bake.md
@@ -1,5 +1,5 @@
 ---
-'@orchestrator-ui/orchestrator-ui-components': patch
+'@orchestrator-ui/orchestrator-ui-components': major
 ---
 
-1479 GroupedTable: The table has a small header with a button "Collapse / Expand", this can now be overridden. csvDownload: Changed the parameters of initiateCsvFileDownload for directly exporting data to CSV. Introduced a utility type to help ensuring that all the props of an object will be of type "string". Typically useful for the csvDownload functionality
+1479 csvDownload (Breaking change): Changed the parameters of initiateCsvFileDownload for directly exporting data to CSV. GroupedTable: The table has a small header with a button "Collapse / Expand", this can now be overridden. Introduced a utility type to help ensuring that all the props of an object will be of type "string". Typically useful for the csvDownload functionality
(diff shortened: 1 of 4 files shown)


packages/orchestrator-ui-components/src/utils/csvDownload.ts
```diff
@@ -26,11 +26,23 @@ function startCsvDownload(csvFileContent: string, fileName: string) {
...
+function totalItemsExceedMaximumItemsForBulkFetching(
+    totalItems?: number | null,
+) {
+    return (totalItems ?? 0) > MAXIMUM_ITEMS_FOR_BULK_FETCHING;
+}
+
+// This function should be used when the data is mapped and ready to be exported
 export function initiateCsvFileDownload<T extends object>(
     data: T[],
+    keyOrder: string[],
```
ricardovdheijden, line 38: Dear reviewer: I am considering doing a major version bump. Officially it should, since this is a breaking change. However this function is not used by our own apps/examples. Please share your thoughts.

packages/orchestrator-ui-components/src/utils/csvDownload.ts
```diff
@@ -47,41 +59,62 @@ export const getCsvFileNameWithDate = (fileNameWithoutExtension: string) => {
...
-                    translationFunction('notAllResultsExportedTitle'),
-                );
-            }
-            initiateCsvFileDownload(dataForExport, filename);
+        const data = dataMapper(fetchResult);
+        const pageInfo = pageInfoMapper(fetchResult);
+
+        if (
+            addToastFunction &&
+            translationFunction &&
```
DutchBen, line 92: In what situation could the translationFunction not exist? And if it doesn't don't we still want to show an error message?

packages/orchestrator-ui-components/src/utils/csvDownload.ts
DutchBen: I'm not sure. It is a breaking change but I think this function is only used in this file and the chances some - as of yet theoretical - external user is using this function are slim. 

If we do follow semver this strict we will end up with versioning like EUI which might be fine, I don't know. Lets use a frontend meeting to discuss our semver strategy a bit more in depth.

DutchBen (approved):

packages/orchestrator-ui-components/src/utils/csvDownload.ts
ricardovdheijden: in a previous commit, they were temporary optional, forgot to clean this up.

.changeset/real-planets-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@orchestrator-ui/orchestrator-ui-components': major
+---
+
+1479 csvDownload (Breaking change): Changed the parameters of initiateCsvFileDownload for directly exporting data to CSV. GroupedTable: The table has a small header with a button "Collapse / Expand", this can now be overridden. Introduced a utility type to help ensuring that all the props of an object will be of type "string". Typically useful for the csvDownload functionality
```
tjeerddie, line 5: this should not be a single line and you should give more context for the breaking change, so the changes can be made by people that use the package

ricardovdheijden: Yes, good one. Will do.

ricardovdheijden: Updated the file manually, but i am not sure the changeset plugin would handle this correctly. By right we should use the cli to enter the changes, which is a single line. The changelog.md file, where the changes are moved to by the changeset-plugin, seems to only have single lines.

PR #1669, opened 2025-01-14
Remove extra parameters causing error on LIR prefix form

Georgi2704:


packages/orchestrator-ui-components/src/components/WfoForms/UserInputForm.tsx
```diff
@@ -407,8 +407,7 @@ function fillPreselection(form: JSONSchema6, router: NextRouter) {
                 }
                 ipPrefixInput.default = `${queryParams.prefix}/${queryParams.prefixlen}`;
                 ipPrefixInput.uniforms.prefixMin = parseInt(
-                    (queryParams.prefix_min as string) ??
```
DutchBen, line 410: Just to double check. Are you sure we are not using prefix_min anywhere?

DutchBen (approved):

PR #1683, opened 2025-01-20
Implement Storybook with Next.js and Vite

Georgi2704:
Known issues:
- Stories that use` <EuiIcon type="..."/>  ` cannot run due to some issue with dynamically imported modules, so I skipped them. It is known issue on Elastic UI repo: https://github.com/elastic/eui/issues/5463 

Could be improved:
- Currently stories are loaded from /storybook/src/stories which means that I moved them to the storybook app. It would be better experience if load them from /packages/orchestrator-ui-components/components next to their original component.I decided for now to not delete the original stories, but keep them in their old locations.

Screenshot 2025-01-22 at 9 04 27

An example setup of preview.tsx and how elastic team is handling the import of icons (potential fix):
https://github.com/elastic/eui/blob/main/packages/eui/.storybook/preview.tsx


apps/storybook/.storybook/main.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * This function is used to resolve the absolute path of a package.
+ * It is needed in projects that use Yarn PnP or are set up within a monorepo.
+ */
+function getAbsolutePath(value: string): any {
+    return dirname(require.resolve(join(value, 'package.json')));
+}
+const config: StorybookConfig = {
+    stories: [
+        '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)',
```
tjeerddie, line 14: this should be the reason the stories did not work outside your folder. we dont have a `../src` anymore and should have been `../packages/` instead

tjeerddie: or did you mean this did not work before creating the `storybook` app folder?

tjeerddie: there is different behavior loading directly from packages folder or loading from the npm package...

apps/storybook/src/stories/WfoNoResults/styles.ts
```diff
@@ -0,0 +1,23 @@
+import { css } from '@emotion/react';
```
tjeerddie, line 1: I believe this file can be deleted

apps/storybook/src/stories/WfoBadges/WfoWebsocketStatusBadge/styles.ts
```diff
@@ -0,0 +1,22 @@
+import { css } from '@emotion/react';
```
tjeerddie, line 1: I believe this file can be deleted

apps/storybook/src/stories/WfoBadges/WfoFailedTasksBadge/styles.ts
```diff
@@ -0,0 +1,10 @@
+import { css } from '@emotion/react';
```
tjeerddie, line 1: I believe this file can be deleted

tjeerddie (approved):

apps/storybook/package.json
```diff
@@ -0,0 +1,46 @@
...
+        "@storybook/test": "^8.5.0",
+        "@types/react": "^18.3.18",
+        "@types/react-dom": "^18.3.5",
+        "@vitejs/plugin-react": "^4.3.4",
+        "eslint": "^8.57.1",
+        "eslint-plugin-react-hooks": "^5.0.0",
+        "eslint-plugin-react-refresh": "^0.4.16",
+        "eslint-plugin-storybook": "^0.10.0",
+        "globals": "^15.14.0",
+        "storybook": "^8.5.0",
```
DutchBen, line 36: I'm not sure it matter but in this context I would expect storybook to not be a dev dependency. Goes for some others aswell

DutchBen (approved):

turbo.json
```diff
@@ -23,8 +23,15 @@
...
-            "outputs": ["dist/**", ".next/**", "!.next/cache/**"]
+            "inputs": ["$TURBO_DEFAULT$", "!**/*.stories.{tsx,jsx,mdx}"],
+            "outputs": [
+                "dist/**",
+                ".next/**",
+                "!.next/cache/**",
+                "storybook-static/**"
+            ]
         },
+        "build:storybook": {},
```
ricardovdheijden, line 34: I think we can remove this line? Or implement a build command in package.json?

PR #1759, opened 2025-02-25
1705 Added prop for custom icon summary card

Georgi2704:


DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoSummary/WfoSummaryCard.tsx
```diff
@@ -57,11 +60,18 @@ export const WfoSummaryCard: FC<WfoSummaryCardProps> = ({
                     iconColor: theme.colors.danger,
                 };
             case SummaryCardStatus.Neutral:
-            default:
                 return {
                     iconType: 'kubernetesPod',
                     iconColor: theme.colors.primary,
                 };
+            case SummaryCardStatus.Custom:
```
DutchBen, line 67: Just double checking. If this case is the same as default is it really neccesary?

Georgi2704: I tried to be more explicit here. Now I think it is a better idea to make the `headerStatus` prop optional. If it's empty it will go straight to the default case.

ricardovdheijden: The default case does not exist, right? `status` at line 49 tells us it is only an enum. Therefore `default` case should never happen. Maybe consider logging/throwing and error when the default case is triggered.

ricardovdheijden (approved):

PR #1813, opened 2025-03-24
Improve summary cards loading states

Georgi2704:
Screenshot 2025-03-24 at 10 49 23

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoSummary/WfoActiveWorkflowsSummaryCard.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoSummary/WfoActiveWorkflowsSummaryCard.tsx
@@ -18,6 +18,7 @@ export const WfoActiveWorkflowsSummaryCard = () => {
     const {
         data: activeWorkflowsSummaryResponse,
         isFetching: activeWorkflowsSummaryIsFetching,
+        isLoading: activeWorkflowsSummaryIsLoading,
     } = useGetProcessListSummaryQuery(activeWorkflowsListSummaryQueryVariables);
 
     return (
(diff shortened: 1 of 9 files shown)


packages/orchestrator-ui-components/src/components/WfoSummary/WfoActiveWorkflowsSummaryCard.tsx
```diff
@@ -18,6 +18,7 @@ export const WfoActiveWorkflowsSummaryCard = () => {
     const {
         data: activeWorkflowsSummaryResponse,
         isFetching: activeWorkflowsSummaryIsFetching,
+        isLoading: activeWorkflowsSummaryIsLoading,
```
DutchBen, line 21: You don't have to rename here right? You could pass isLoading to isLoading in line 37 directly?

packages/orchestrator-ui-components/src/components/WfoSummary/WfoFailedTasksSummaryCard.tsx
```diff
@@ -14,6 +14,7 @@ export const WfoFailedTasksSummaryCard = () => {
     const {
         data: failedTasksSummaryResponse,
         isFetching: failedTasksSummaryIsFetching,
+        isLoading: failedTasksSummaryIsLoading,
```
DutchBen, line 17: Same as above

packages/orchestrator-ui-components/src/components/WfoSummary/WfoActiveWorkflowsSummaryCard.tsx
Georgi2704: They were renamed before, so I decided to stick to the convention. Maybe @ricardovdheijden has an opinion ?

ricardovdheijden: Yes I renamed them before when they were all together in one file. But now since there is one api call per component, it does look nicer to just stick with `isLoading` and `isFetching` etc

ricardovdheijden (approved):

PR #1872, opened 2025-04-18
Remove inUseByRelations field from query SubscriptionDropdownOptions

Mark90:
Related orchestrator-core PR: https://github.com/workfloworchestrator/orchestrator-core/pull/922

### DIFF
--- a/packages/orchestrator-ui-components/src/types/types.ts
+++ b/packages/orchestrator-ui-components/src/types/types.ts
@@ -62,13 +62,10 @@ export type ProductBlockInstance = {
     subscription: Pick<Subscription, 'subscriptionId' | 'description'>;
 };
 
-export type ProductBlockInstanceForDropdown = {
-    id: number;
-    subscriptionInstanceId: string;
-    parent: Nullable<number>;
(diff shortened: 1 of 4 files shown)


packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -62,6 +62,14 @@ export type ProductBlockInstance = {
...
 };
 
+export type ProductBlockInstanceForDropdown = {
+    id: number;
+    subscriptionInstanceId: string;
+    parent: Nullable<number>;
+    productBlockInstanceValues: FieldValue[];
+    subscription: Pick<Subscription, 'subscriptionId' | 'description'>;
+};
+
```
ricardovdheijden, line 72: Is this the same type as ProductBlockInstance, but without the inUseByRelations prop? If yes, it might be better to use the Omit utility type

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/types/types.ts
Mark90: Done!

PR #1892, opened 2025-04-30
1802 npm audit

DutchBen:
Updates all npm packages as far as they can go. See https://github.com/workfloworchestrator/orchestrator-ui-library/issues/1903 for packages I wasn't able to update.


package.json
```diff
@@ -17,8 +17,10 @@
     },
     "devDependencies": {
         "@changesets/cli": "^2.27.1",
-        "@turbo/gen": "^2.2.3",
+        "@trivago/prettier-plugin-sort-imports": "^5.2.2",
+        "@turbo/gen": "^2.5.2",
         "husky": "^9.0.11",
+        "prettier": "^3.5.3",
```
ricardovdheijden, line 23: Why would you need prettier on this highest level? If still needed, maybe consider importing the "@orchestrator-ui/eslint-config-custom" package containing Prettier and Trivago.

packages/orchestrator-ui-components/package.json
```diff
@@ -30,26 +30,26 @@
...
         "dev": "npm run build -- --watch"
     },
     "dependencies": {
-        "@elastic/eui": "^97.0.0",
+        "@elastic/eui": "^101.3.0",
         "@emotion/css": "^11.11.2",
         "@emotion/react": "^11.11.4",
         "@rtk-query/graphql-request-base-query": "^2.3.1",
-        "eslint": "^8.57.0",
+        "@trivago/prettier-plugin-sort-imports": "^5.2.2",
```
ricardovdheijden, line 41: This one comes via the "@orchestrator-ui/eslint-config-custom" package

packages/orchestrator-ui-components/package.json
```diff
@@ -73,6 +73,7 @@
         "@types/object-hash": "^3.0.6",
         "@types/scroll-into-view": "^1.16.3",
         "babel-jest": "^29.7.0",
+        "eslint": "^9.25.1",
```
ricardovdheijden, line 76: eslint is part of "@orchestrator-ui/eslint-config-custom" dependency. No need to import extra, it would introduce more problems if we fail te keep all versions in sync

package.json
DutchBen: No you're right. They can be removed here.

packages/orchestrator-ui-components/package.json
DutchBen: removed here

packages/orchestrator-ui-components/package.json
DutchBen: removed

ricardovdheijden (approved):

Georgi2704 (approved):

PR #1896, opened 2025-05-01
Replace hardcoded query params with constants

Georgi2704:


wouter1975 (approved):

packages/orchestrator-ui-components/src/components/WfoSummary/WfoLatestOutOfSyncSubscriptionSummaryCard.tsx
```diff
@@ -23,6 +25,13 @@ export const WfoLatestOutOfSyncSubscriptionSummaryCard = () => {
         outOfSyncSubscriptionsListSummaryQueryVariables,
     );
 
+    const queryParams = {
+        [WfoQueryParams.ACTIVE_TAB]: WfoSubscriptionListTab.ALL,
+        [WfoQueryParams.SORT_BY]: 'field-startDate_order-ASC',
+        [WfoQueryParams.QUERY_STRING]:
+            'status:(provisioning|active) insync:false',
```
mrijk, line 32: I probably don't have enough background on this issue, but I don't see how this really helps. Looks like just key/value pairs (which the key as enums and thus adding additional overhead in javascript) and the value kan still be any string?

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoSummary/WfoLatestOutOfSyncSubscriptionSummaryCard.tsx
DutchBen: At least if avoids hardcoded strings for keys. It would be beter to enforce the shape of queryParams aswell so we can enforce the key to always be a WfoQueryParam as it expects when the getUrlWithQueryParams function gets called.
Side question. Have you looked at the solution Ricardo linked in his ticket and if/why have you decidced against that? https://www.npmjs.com/package/use-query-params#encodequeryparams

Georgi2704: Yes the value can be any string, but the key is stored as a constant. For cases such as we decide to change the key 'filterBy' to 'filter' in the backend. Otherwise there will be magic strings everywhere.

Georgi2704: @DutchBen because I found out that new URLSearchParams() does what 

```
const encodedQuery = encodeQueryParams({ foo: NumberParam }, { foo });
const link = `/?${stringify(encodedQuery)}`;
```

does already without the need of a new dependency

packages/orchestrator-ui-components/src/utils/getQueryParams.ts
```diff
@@ -0,0 +1,16 @@
...
+    PAGE = 'page',
+    PAGE_SIZE = 'pageSize',
+    SORT_BY = 'sortBy',
+    FILTER_BY = 'filterBy',
+    QUERY_STRING = 'queryString',
+}
+
+export const getUrlWithQueryParams = (
+    url: string,
+    params: Partial<Record<WfoQueryParams, string>>,
```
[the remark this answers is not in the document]
Georgi2704: Because otherwise you always need to include all the query params specified in WfoQueryParams. Sometimes you need only  1 or 2 of them.

PR #1897, opened 2025-05-01
removes conditional render option to edit description on wf/task metad…

wouter1975:
removes conditional render option to edit description on wf/task metadata page, requires orch-core 4.0.0

### DIFF
--- a/packages/orchestrator-ui-components/src/pages/metadata/WfoWorkflowsPage.tsx
+++ b/packages/orchestrator-ui-components/src/pages/metadata/WfoWorkflowsPage.tsx
@@ -131,19 +131,34 @@ export const WfoWorkflowsPage = () => {
             columnType: ColumnType.DATA,
             label: t('description'),
             width: '700px',
-            renderData: (value, row) =>
-                value ? (
-                    <WfoMetadataDescriptionField
-                        onSave={(updatedNote) =>
-                            updateWorkflow({
-                                id: row.workflowId,
-                                description: updatedNote,
-                            })
-                        }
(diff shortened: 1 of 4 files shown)


packages/orchestrator-ui-components/src/pages/metadata/WfoWorkflowsPage.tsx
```diff
@@ -131,19 +131,34 @@ export const WfoWorkflowsPage = () => {
...
+                        updateWorkflow({
+                            id: row.workflowId,
+                            description: updatedNote,
+                        })
+                    }
+                    description={value}
+                />
+            ),
         },
+        // description: {
```
DutchBen, line 146: Either uncomment this or remove it. Is there a reason to keep it?

packages/orchestrator-ui-components/src/types/types.ts
```diff
@@ -258,7 +258,7 @@ export interface Step {
 export interface WorkflowDefinition {
     workflowId: string;
     name: string;
-    description?: string;
+    description: string;
```
ricardovdheijden, line 261: By right, this is a breaking change. I also saw that this type is used in multiple other places:
image

Did you verify all of them? Same goes for the TaskDefinition type below

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/types/types.ts
wouter1975: Yes, should be fine

PR #2012, opened 2025-06-23
Fix query error in WfoSubscriptionNoteEdit by using existing data

tjeerddie:
the component was re-fetching data without query variables which resulted in query errors `const endpointName = useQuery().endpointName;` but was also confused why the data is refetched

- change WfoSubscriptionNoteEdit to use existing data
- clean WfoSubscriptionsList and move the subscription list fetch above the table columns so the list can be used in WfoSubscriptionNoteEdit. (also cleaned up the html)
- change subscriptionResultMappers to work with response or undefined and returns empty list when undefined.

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoInlineNoteEdit/WfoSubscriptionNoteEdit.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoInlineNoteEdit/WfoSubscriptionNoteEdit.tsx
@@ -1,41 +1,32 @@
 import type { FC } from 'react';
 import React from 'react';
 
-import { ApiResult, SubscriptionListResponse, UseQuery } from '@/rtk';
 import { useStartProcessMutation } from '@/rtk/endpoints/forms';
 import { useUpdateSubscriptionNoteOptimisticMutation } from '@/rtk/endpoints/subscriptionListMutation';
-import { Subscription } from '@/types';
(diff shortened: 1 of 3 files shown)


packages/orchestrator-ui-components/src/components/WfoInlineNoteEdit/WfoSubscriptionNoteEdit.tsx
```diff
@@ -1,35 +1,32 @@
...
-        selectFromResult: (result: ApiResult<SubscriptionListResponse>) => ({
-            selectedItem: result?.data?.subscriptions?.find(
-                (sub) => sub.subscriptionId === subscriptionId,
-            ),
-        }),
-    });
-    const endpointName = useQuery().endpointName;
+    const selectedItem = data.find(
+        (sub) => sub.subscriptionId === subscriptionId,
+    );
```
ricardovdheijden, line 29: Can't we just pass the current row into this component? Now we pass all data and we need to perform a `find` to retrieve it again

ricardovdheijden (approved):

packages/orchestrator-ui-components/src/components/WfoInlineNoteEdit/WfoSubscriptionNoteEdit.tsx
tjeerddie: yup 🤦‍♂️

PR #2020, opened 2025-06-25
1994 Adds classnames to key and value cells

ricardovdheijden:


Georgi2704 (approved):

packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionProductBlock/WfoProductBlockKeyValueRow.tsx
```diff
@@ -9,14 +9,18 @@ import { camelToHuman } from '@/utils';
 
 import { getStyles } from './styles';
 
+export const KEY_CELL_CLASS_NAME = 'key-cell';
+export const VALUE_CELL_CLASS_NAME = 'value-cell';
+
 export type WfoProductBlockKeyValueRowProps = {
     fieldValue: FieldValue | RenderableFieldValue;
     allFieldValues: FieldValue[] | RenderableFieldValue[];
+    className?: string;
```
DutchBen, line 18: Is this being used. In the surf app you only use KEY_CELL_CLASS_NAME right? Is this emotion magic or is it meant to be passed in some way?

ricardovdheijden: yes, `className` is needed to pass in the css prop to custom components. In this case the `WfoProductBlockKeyValueRow` is a custom component. So indeed, you will see me adding `className` here and applying `css` on the consuming side. This is part of emotion-css by the way

PR #2043, opened 2025-07-02
Disable workflow starts that are not allowed for the current user

### LINKED ISSUE #2013: [Feature]: Add Role based access control to start workflows - Implement updated endpoint in UI
Depends on backend change: https://github.com/workfloworchestrator/orchestrator-core/issues/821

In several places the update endpoints for running workflow/tasks need to updated in the FE.

- [ ] Reflect the isAllowed state in the Start menu

is workflow has isAllowed = false, make the menu item (shown by the product name) not-clickable, and present the item in a disabled state (grey text), consider adding a lock to indicate that the user has insufficient rights and add on-hoover stating the user is not allowed to execute this workflow

kvklink:
Depends on a backend change.
Closes #2013 

image
When a user is not allowed to start a workflow according to the backend, the options are grayed out.

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartWorkflowComboBox.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartWorkflowComboBox.tsx
@@ -28,6 +28,7 @@ export const WfoStartWorkflowButtonComboBox = () => {
                 productId: option.productId,
             },
             label: option.productName || '',
+            disabled: !option.isAllowed,
         }));
 
     const handleOptionChange = async (selectedProduct: StartComboBoxOption) => {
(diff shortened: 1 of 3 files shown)


kvklink:
FYI: This will depend on a backend change so the version number will have to increase accordingly on both back- and front-end with the respective required backend version here. I just don't know how to do that...

packages/orchestrator-ui-components/src/components/WfoStartButton/WfoStartWorkflowComboBox.tsx
```diff
@@ -28,6 +28,7 @@ export const WfoStartWorkflowButtonComboBox = () => {
                 productId: option.productId,
             },
             label: option.productName || '',
+            disabled: !option.isAllowed,
```
DutchBen, line 31: This makes it so it defaults to disabled correct? I would prefer it to 
```suggestion
            disabled: option.isAllowed && option.isAllowed === false,
```

DutchBen (approved):

PR #2117, opened 2025-08-18
2067 select field

DutchBen:


Georgi2704 (approved):

packages/orchestrator-ui-components/src/components/WfoPydanticForm/WfoPydanticForm.tsx
```diff
@@ -168,6 +171,21 @@ export const useWfoPydanticFormConfig = () => {
...
+                        Element: WfoDropdown,
+                        isControlledElement: true,
+                    },
+                    matcher(field) {
+                        // We are looking for a single value from a set list of options. With more than 3 options use a dropdown.
+                        return (
+                            field.type === PydanticFormFieldType.STRING &&
+                            field.options.length > 0 &&
+                            field.options.length > 3
+                        );
```
tvdven, line 186: Wouldn't it be better to perform the following check to avoid runtime erros when options is undefined or null:

```
return (
    field.type === PydanticFormFieldType.STRING &&
    Array.IsArray(field.options) &&
    field.options.length > 3
);
```

Or use chaining like:

```
return (
    field.type === PydanticFormFieldType.STRING &&
    field.options?.length > 3
);
```

tvdven (commented):
One small comment

packages/orchestrator-ui-components/src/components/WfoPydanticForm/WfoPydanticForm.tsx
DutchBen: Fixed. Also remove this matcher in the mean time

PR #2157, opened 2025-09-08
Make Pydantic form row use unique IDs

Georgi2704:


packages/orchestrator-ui-components/src/components/WfoPydanticForm/Row.tsx
```diff
@@ -29,7 +29,7 @@ export const Row: RowRenderComponent = ({
 
     return (
         <EuiFormRow
-            data-testid="pydantic-form-row"
+            data-testid={getDataTestId('pydantic-form-row', title)}
```
DutchBen, line 32: you should ask @Floris if this works for him or if he wants the rows number by their position. So `pydantic-form-row-0` for the first, `pydantic-form-row-1` for the second and so on

DutchBen (approved):

PR #2181, opened 2025-09-23
Add reconcile to Subscription Actions

Georgi2704:


packages/orchestrator-ui-components/src/components/WfoSubscription/WfoSubscriptionActions/WfoSubscriptionActions.tsx
```diff
@@ -116,35 +117,7 @@ export const WfoSubscriptionActions: FC<WfoSubscriptionActionsProps> = ({
         };
 
         const tooltipIt = (actionItem: React.ReactNode) => {
-            /**
```
DutchBen, line 119: I would like to keep this documentation in here. It's a bit of a mindfuck if you run into this without explanation.

PR #2193, opened 2025-09-25
Use monaco-editor in WfoJsonCodeBlock

RichieB2B:
We have many large dicts in our workflow states. Very useful for debugging but it is a pain to find the right piece of information. I replaced the `EuiCodeBlock` with `monaco-editor` in read-only mode. This gives the benefit of collapsable blocks, sticky scrolling and code-only zoom. The only thing that is missing, is a copy-all button. But Cmd+A + Cmd+C is just as easy.

image001

### DIFF
--- a/packages/orchestrator-ui-components/src/components/index.ts
+++ b/packages/orchestrator-ui-components/src/components/index.ts
@@ -36,9 +36,6 @@ export * from './WfoInlineNoteEdit';
 export * from './WfoTableCodeBlock';
 export * from './WfoInlineEdit';
 export * from './WfoPydanticForm';
-<<<<<<< HEAD
 export * from './WfoSearchPage';
 export * from './WfoAgent';
-=======
 export * from './WfoMonacoCodeBlock';
->>>>>>> eed74dca (Make monaco editor a separate component)


packages/orchestrator-ui-components/src/components/WfoJsonCodeBlock/WfoJsonCodeBlock.tsx
```diff
@@ -1,8 +1,10 @@
 import React, { FC } from 'react';
```
DutchBen, line 1:
```suggestion
import React, { FC, useState } from 'react';
```

packages/orchestrator-ui-components/src/components/WfoJsonCodeBlock/WfoJsonCodeBlock.tsx
```diff
@@ -21,19 +23,39 @@ export const WfoJsonCodeBlock: FC<WfoJsonCodeBlockProps> = ({
     data,
     isBasicStyle = false,
 }) => {
+    const { isDarkThemeActive } = useOrchestratorTheme();
     const { euiCodeBlockStyle, euiBasicCodeBlockStyle } =
         useWithOrchestratorTheme(getStyles);
 
     const json = JSON.stringify(data, null, 4);
 
+    const [editorHeight, setEditorHeight] = React.useState(0);
```
DutchBen, line 32:
```suggestion
    const [editorHeight, setEditorHeight] = useState(0);
```

pboers1988:
@RichieB2B deferring to @DutchBen

PR #2204, opened 2025-09-30
Added link for workflow in toast

Georgi2704:


packages/orchestrator-ui-components/src/utils/getToastMessage.ts
```diff
@@ -1,10 +1,12 @@
+import { ReactNode } from 'react';
+
 import { ToastTypes } from '../types';
 import type { Toast } from '../types';
 
 export const getToastMessage = (
     type: ToastTypes,
-    text: string, // We use string here instead of Toast['text'] because we want to prevent passing in react component because they trigger an "unsynchronizable values in payload detected" error',
```
DutchBen, line 6: Does the error this comment mentions not get triggered anymore?

packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetail.tsx
```diff
@@ -105,6 +117,7 @@ export const WfoProcessDetail = ({
     const [retryProcess] = useRetryProcessMutation();
     const [deleteProcess] = useDeleteProcessMutation();
     const [abortProcess] = useAbortProcessMutation();
+    const hasPrev = useHasPreviousRoute();
```
DutchBen, line 120:
```suggestion
    const hasPreviousRoute = useHasPreviousRoute();
```

packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetail.tsx
```diff
@@ -177,7 +190,12 @@ export const WfoProcessDetail = ({
                 if (processDetail?.processId) {
                     abortProcess({ processId: processDetail.processId });
                 }
-                router.push(processIsTask ? PATH_TASKS : PATH_WORKFLOWS);
+
+                if (hasPrev) {
```
DutchBen, line 194:
```suggestion
                if (hasPreviousRoute) {
```

packages/orchestrator-ui-components/src/pages/processes/WfoProcessDetail.tsx
```diff
@@ -88,6 +88,18 @@ interface ProcessDetailProps {
     hasError?: boolean;
 }
 
+function useHasPreviousRoute() {
```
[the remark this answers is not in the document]
Georgi2704: @DutchBen Its not from this issue, but it was bothering me that it always redirected you to the workflow list page after aborting. So coming from the Subscription page, then workflow detail, then workflow I had to go 2 times back so i can return where i came from. I also mentioned it to Wouter yesterday. However here may be not the best place for it, I can move it in some generic utils file.

PR #2304, opened 2025-11-06
render tool calls properly

timfdev:
depends on [#1149](https://github.com/workfloworchestrator/orchestrator-core/pull/1149)


package.json
```diff
@@ -17,8 +17,10 @@
     },
     "devDependencies": {
         "@changesets/cli": "^2.27.1",
```
timfdev: Adding these fixed my missing package issues, they seem to be part of the submodule but somehow that just doesn't work well when I try a clean install.

pboers1988 (approved):

PR #2317, opened 2025-11-12
Update pydantic-forms to v0.10.0

Georgi2704:


apps/wfo-ui
wouterhhuisman: Is this file needed?

DutchBen (approved):

PR #2508, opened 2026-04-14
add explanation for cron input

wouterhhuisman:
Add explanation on valid input of cron input form


DutchBen (approved):

DutchBen:
The schedule form is served form the backend now so I think this won't work anymore.

wouterhhuisman:
Is superseeded by the backend driven form to add a schedule.

PR #2522, opened 2026-04-21
Deploy script

dev-elie:
## Background

At Virge we use the WFO for various things, but as of right now it's a bit of a pain when upgrading and redeploying the WFO-UI. Vercel for example doesn't play along with the whole submodules thing, so I made a deploy script that automates that pain away. The script creates a deployable version by checking out the selected package tag, materializing `apps/wfo-ui` as regular tracked files, and preparing a self-contained deploy branch.

## What this PR adds

Adds a `npm run deploy` script for creating deploy branches from published `@orchestrator-ui/orchestrator-ui-components` tags.

Why? The published UI package version and the deployable example app are not represented as a single branch by default. The new script turns that into a repeatable process; it checks out the selected package tag, converts the `apps/wfo-ui` submodule into regular tracked files, and produces a self-contained deploy branch that Vercel (or any other host) can work with directly.

## What changed

- `scripts/deploy.mjs` interactive deploy entrypoint
- `npm run deploy` / `npm run deploy:quick` added to `package.json`
- `@inquirer/prompts` added as a dev dependency
- README updated with deploy workflow documentation

## Deploy flow (`npm run deploy` )

- Select the package version tag to deploy
- Choose a safe push remote (refuses pushes to the official repo); can create a fork via GitHub CLI if none exists
- Pick or create the output branch
- Optionally add `@copilotkit/runtime` when the selected tag needs it
- Materializes `apps/wfo-ui` submodule as regular tracked files
- Pushes the self-contained deploy branch

A `--quick` / `-q` flag skips all prompts and takes safe defaults (latest tag, first safe remote, suggested branch name, no force push).


README.md
```diff
@@ -109,3 +109,21 @@ npx storybook dev
 ```
 
 Story book can be inspected on [http://localhost:61834/](http://localhost:61834/).
+
+## Create a deploy branch
+
+Vercel and similar hosts don't play well with monorepos, so use the deploy script to produce a self-contained branch from a published `@orchestrator-ui/orchestrator-ui-components` tag the submodule gets materialized as regular tracked files and everything is pushed in one shot.
```
acidjunk: Vercel can deploy fine from a mono repo. But when other people want to do that they can't just add Vercel to this repo (that's how Vercel works; it get access to a repo and then "autodeploys").

It's a long build (12 minutes and you need 8Gb of Ram). It doesn't understand the sub module though...

So at Virge we now have a fork and inside that fork we have a version of the wfo-ui reference app inside one repo (truly a mono repo 🙈 ). This scripts makes it easy to align our fork with the official upstream repo's AND still have an easy way to edit stuff when the need arises.

package-lock.json
acidjunk, line 1: @dev-elie I think this is an error.

Hypothese: you didn't have the SURF repo checked out (GPL maintainer in general will not have access to it). So when you then install packages it will remove all `wfo-ui-surf` references. I think other GPL maintainerer will have the same experience; so this makes it almost impossible to contribute (to this repo), when you not work for SURF directly.

acidjunk: @DutchBen : tagging you here, because we discussed this exact issue a while ago.

acidjunk (changes requested):
Try to fix package-lock and make the docs clear regarding the fork construction. People will no run the script from this repo, but are expected to run it from their own fork.
(I know; it feels complex)

dev-elie:
Package-lock should be better like this @acidjunk , I checked out the SURF repo and installed again. And i also updated the README

DutchBen:
This looks good and useful for your current use case. I would just like to add - for reference for others aswell - that this is not the recommended way of deploying workflow orchestrator frontends. What we do at surf is we use the component library published to NPM (from this library repo) inside our custom surf app. 
https://www.npmjs.com/package/@orchestrator-ui/orchestrator-ui-components

So what is recommended:
- Create/Fork an app based on https://github.com/workfloworchestrator/example-orchestrator-ui/
- Add customisations and add extra pages
- We have a gitlab repository that holds the surf version of the app and release from there.

We don't use this repository (https://github.com/workfloworchestrator/orchestrator-ui-library/) to deploy anything other than publishing to NPM.

Contributing to this repository and having these contributions released to NPM so they end up in https://www.npmjs.com/package/@orchestrator-ui/orchestrator-ui-components is currently complicated since the package-lock file is polluted with surf app related components because of the development setup. Which is setup the way you are using it. That's a known issue that we have on the roadmap and the we will work on together with Verge. I believe there is a meeting planned about this already.

DutchBen:
This is useful for some usecases but not the recommended way of deploying so we are not mergin it into the main branch.

PR #2559, opened 2026-05-19
wip: 94 filter builder

DutchBen:
Implements React query builder component to use for searching on data pages like the subsciption list or workflow list page.


Georgi2704:
Added into: https://github.com/workfloworchestrator/orchestrator-ui-library/pull/2627

PR #2562, opened 2026-05-20
Small cosmetic improvements

wouterhhuisman:
- Added tooltip on expand row to detail view icon in wfotable and added EUI highlighting
- Changed colour of subscription delta icon to grey (textdisabled)
- Hide the expand icon next to status badge in metadata/products
- By default show type and status in metadata/products

image
image
image


packages/orchestrator-ui-components/src/utils/getDefaultTableConfig.ts
```diff
@@ -42,7 +42,7 @@ export const getDefaultTableConfig = <T>(storageKey: string) => {
       return getTableConfig<T>(resourceTypeColumns as (keyof T)[]);
     }
     case METADATA_PRODUCT_TABLE_LOCAL_STORAGE_KEY: {
-      const productColumns: (keyof ProductDefinition)[] = ['productId', 'productType', 'status', 'createdAt'];
+      const productColumns: (keyof ProductDefinition)[] = ['productId', 'createdAt'];
```
DutchBen, line 45: Why are these removed?

wouterhhuisman: I think these columns should be visible by default? Don't you agree?

PR #2572, opened 2026-06-01
Update NPM packages

Georgi2704:


.npmrc
```diff
@@ -0,0 +1,4 @@
+# Only install package versions published more than 14 days ago.
+# Gives the ecosystem time to detect malicious or compromised releases.
+# Requires npm >= 11.15.0.
+min-release-age=14
```
surf-anieuwland: As Mark also said earlier, in central discussions we concluded 14 is too long. Often vulnerabilities are already detected within a day. We thought 2 days would be better.

surf-anieuwland (approved):

.npmrc
Georgi2704:
```suggestion
min-release-age=2
```

DutchBen (approved):

PR #2598, opened 2026-06-17
Show error toast on tasks that return 412 error code

Georgi2704:
Screenshot 2026-06-17 at 15 38 39

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoPydanticForm/WfoPydanticForm.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoPydanticForm/WfoPydanticForm.tsx
@@ -1,16 +1,19 @@
 import React, { useCallback, useMemo } from 'react';
 
 import _ from 'lodash';
+import { useTranslations } from 'next-intl';
 import { useRouter } from 'next/router';
 import type { PydanticFormApiProvider } from 'pydantic-forms';
 import { PydanticForm } from 'pydantic-forms';
 
 import { PATH_TASKS, PATH_WORKFLOWS } from '@/components';
 import { Footer } from '@/components/WfoPydanticForm/Footer';
+import { useShowToastMessage } from '@/hooks';
 import { useGetPydanticFormsConfig } from '@/hooks/useGetPydanticFormsConfig';
 import { StartWorkflowPayload } from '@/pages/processes/WfoStartProcessPage';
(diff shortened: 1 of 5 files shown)


packages/orchestrator-ui-components/src/components/WfoPydanticForm/WfoPydanticForm.tsx
```diff
@@ -64,6 +69,23 @@ export const WfoPydanticForm = ({ processName, startProcessPayload, isTask }: Wf
...
                   status: error.status,
                 });
+              } else if (error.status === HttpStatus.PreconditionFailed) {
+                const detail = typeof error.data.detail === 'string' ? error.data.detail : '';
+                showToastMessage(
+                  ToastTypes.ERROR,
+                  detail || t('preconditionFailedFallback'),
+                  t('preconditionFailedTitle'),
+                );
+                router.replace(isTask ? PATH_TASKS : PATH_WORKFLOWS);
```
DutchBen, line 79: I don't think you need to redirect here right? Show the toast and leave the form untouched so you could retry it

packages/orchestrator-ui-components/src/components/WfoPydanticForm/WfoPydanticForm.tsx
```diff
@@ -64,6 +69,23 @@ export const WfoPydanticForm = ({ processName, startProcessPayload, isTask }: Wf
...
+                resolve({
```
DutchBen, line 80: Also no need to resolve anything here I think. The toast already shows the message. Or just a resolve({status: error})?

DutchBen (approved):
The redirect needs to be removed before merging

packages/orchestrator-ui-components/src/components/WfoPydanticForm/WfoPydanticForm.tsx
Georgi2704: the behaviour is the same in Service Tickets, where it redirects you back.

packages/orchestrator-ui-components/src/components/WfoPydanticForm/WfoPydanticForm.tsx
Georgi2704: Yes, otherwise it will throw error in the console.

wouterhhuisman (approved):

PR #2619, opened 2026-07-02
Workflow user guides

Georgi2704:


packages/orchestrator-ui-components/src/components/WfoWorkflowUserGuide/WfoFormWithUserGuide.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { useGetOrchestratorConfig } from '@/hooks';
+
+import { WfoWorkflowGuideExpandablePanel } from './WfoWorkflowGuideExpandablePanel';
+
+interface WfoFormWithUserGuideProps {
+  workflowName?: string;
+  children: ReactNode;
+}
+
+export const WfoFormWithUserGuide = ({ workflowName, children }: WfoFormWithUserGuideProps) => {
```
DutchBen: This could be named better. It's not always a form that is wrapped with a userguide here right? PageWithUserGuide or something? ComponentWithUserGuide?

packages/orchestrator-ui-components/src/components/WfoWorkflowUserGuide/WfoFormWithUserGuide.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [isExpanded, setIsExpanded] = useState(false);
+  const isStacked = useIsWithinBreakpoints(['xs', 's', 'm', 'l']);
```
DutchBen: I'm not so sure about the responsive parts. We don't do it anywhere else and it does add some overhead and complexity. On the other hand if it's there and it works..

DutchBen (approved):

packages/orchestrator-ui-components/src/components/WfoWorkflowUserGuide/WfoFormWithUserGuide.tsx
Georgi2704: useIsWithinBreakpoints its a hook from elasticUI that i found out about. Seems to work nicely.

packages/orchestrator-ui-components/src/components/WfoWorkflowUserGuide/WfoFormWithUserGuide.tsx
Georgi2704: done

tjeerddie (approved):

PR #2658, opened 2026-07-31
fix(scheduled-tasks): post/delete to /api/schedules/ with trailing slash

### LINKED ISSUE #2649: Creating a scheduled task should post to /api/schedules/ with a trailing slash
The DELETE/POST schedules endpoints have a trailing slash:

Image

But the UI calls the endpoint without a trailing slash in `packages/orchestrator-ui-components/src/rtk/endpoints/metadata/scheduledTasks.ts` which causes unnecessary redirects. (reported in https://github.com/workfloworchestrator/orchestrator-core/issues/1794)

The UI should call the endpoint with a trailing slash.

torkashvand:
## Summary

Fixes #2649. Scheduled-task create (`POST`) and delete (`DELETE`) requests now target the canonical `/api/schedules/` endpoint with a trailing slash, eliminating unnecessary HTTP redirects.

## Root cause

`METADATA_SCHEDULES_ENDPOINT` was defined as `'schedules'` (no trailing slash). Both `createScheduledTask` and `deleteScheduledTask` use it as a bare collection URL. The backend mounts these operations under `/schedules/` (router `POST /` and `DELETE /`), so the server issued a 307 redirect from `/api/schedules` to `/api/schedules/`.

Both are collection endpoints that carry a request body (not `/{id}` item routes), so a trailing slash is correct and produces no double slash.

## Changes

- `constants.ts`: `METADATA_SCHEDULES_ENDPOINT` `'schedules'` -> `'schedules/'`.
- `scheduledTasks.ts`: exported `scheduledTasksApi` to enable behavioural testing (no runtime change).
- `scheduledTasks.spec.ts` (new): dispatches both mutations against a real store with mocked fetch, asserting the canonical URL, method, body, and absence of a double slash.
- Added a `patch` changeset.

The read path (GraphQL query) is unaffected.

## Tests

- `jest` full package suite: 39 suites / 259 tests pass.
- `tsc`, `eslint`, and `tsup` build all pass.
- Verified the new spec fails against the old `'schedules'` value and passes with the fix.


DutchBen:
Here me out on the test that was added. I wonder if we should add it for a change this small and am inclined to see it as scope creep. 

The only change/fix introduced is the adding of the '/'. If you add a test for this you would ideally only test that. The test that was added checks for '//schedule' or 'schedule//' preventing future changes the reintroduce the extra '/' . I would say it's more effective to - if we introduce a test - to just test the value of the schedules constant to see if it actually is 'schedule/'. That way we also guard against 'XXX' for example. 

I recognize that test coverage of api calls is missing and it's takes less effort to add it with current ai tools. This pr introduces boilerplate for setting up a test, a lot of that is general and would be required/used for other future tests aswell. My argument would be that if we want to improve test coverage of this code we should do that in a dedicated ticket so we can think about and discuss what te best setup would be. That would also be true for the guarding against double '//' in urls, that feels like a more generic check that should live somewhere more global.

DutchBen (approved):

packages/orchestrator-ui-components/src/rtk/endpoints/metadata/scheduledTasks.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    expect(url).toBe(`${API_BASE_URL}/schedules/`);
+    expect(url).not.toContain('//schedules');
+    expect(url).not.toContain('schedules//');
+    expect(method).toBe('DELETE');
+    expect(JSON.parse(body)).toEqual({
+      scheduled_type: 'delete',
+      workflow_id: 'workflow-id',
+      schedule_id: 'schedule-id',
+    });
+  });
```
Mark90: I agree that these tests don't add a lot of value, in that they're mostly testing whether the filled in strings (`schedules/` and `DELETE`) don't change. Might be better to create a ticket for improving test coverage and defining for this project what is considered valuable tests.

Mark90 (approved):

Mark90 (approved):

PR #2660, opened 2026-08-03
2578: Step form fix on 7.* branch

DutchBen:


packages/orchestrator-ui-components/src/components/WfoWorkflowSteps/WfoStep/WfoStep.tsx
Mark90, line 187: How does this fix it?

And what is the advise to people stuck on 7.7.0 as in the bug ticket?

DutchBen: There will be a 7.7.1 release with this bugfix and the advise would be to upgrade to that

PR #2677, opened 2026-08-13
Change hardcoded colors to theme colors

tjeerddie:
- fixes colors in dark mode

Not sure if the theme colors are fine or we want the actual hardcoded colors added to the theme


Georgi2704:
Screenshot 2026-08-13 at 14 15 35

@tjeerddie 
Seems way too dark for me now. The old one looks better to me. I added a switch `isDarkModeActive` so we can get best of both worlds.

wouterhhuisman (approved):

Georgi2704 (approved):

PR #2687, opened 2026-08-18
191 hide full path search

Georgi2704:
Screenshot 2026-08-18 at 12 37 07
Screenshot 2026-08-18 at 12 37 16

### DIFF
--- a/packages/orchestrator-ui-components/src/components/WfoTable/WfoStructuredSearchTable/WfoFieldSelector.tsx
+++ b/packages/orchestrator-ui-components/src/components/WfoTable/WfoStructuredSearchTable/WfoFieldSelector.tsx
@@ -44,15 +44,15 @@ export const WfoFieldSelector = ({ handleOnChange, disabled, rule, context }: Fi
     label: path,
   });
 
+  const isSelectablePath = (path: string) => advancedNestedSearch || !path.includes('.');
+
   const getOptionsFromPathInfo = (pathInfos: PathInfo[]): EuiComboBoxOptionOption<string>[] => {
     const pathOptions: EuiComboBoxOptionOption<string>[] = [];
 
     pathInfos.forEach((pathInfo) => {
-      [pathInfo.path, ...(pathInfo.availablePaths ?? [])]
-        .filter((path) => advancedNestedSearch || !path.includes('.'))
-        .forEach((path) => {
-          pathOptions.push(getOption(path));
-        });
+      [pathInfo.path, ...(pathInfo.availablePaths ?? [])].filter(isSelectablePath).forEach((path) => {
+        pathOptions.push(getOption(path));
+      });
     });
     return (
       pathOptions.length > 0 ? pathOptions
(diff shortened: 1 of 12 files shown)


packages/orchestrator-ui-components/src/components/WfoTable/WfoTable/WfoTable.tsx
```diff
@@ -214,7 +219,16 @@ export const WfoTable = <T extends object>({
               />
             </thead>
           }
-          {dataLength === 0 ?
+          {showLoadingSkeleton ?
```
tjeerddie: I would put this in a function, because its to many ifs/logic in the html:
```
const noData = dataLength === 0
const ShowRows = () => {
    if (showLoadingSkeleton) {
        return <SkeletonRows />
    }

    if (noData) {
        return <NoData />
    }

    if (isVirtualized && height ) {
        return <Virtualized?? />
    }

    return <Table />
}
```

tjeerddie (approved):

packages/orchestrator-ui-components/src/components/WfoTable/WfoStructuredSearchTable/WfoStructuredSearchTable.tsx
```diff
@@ -218,6 +237,7 @@ export const WfoStructuredSearchTable = <T extends object>({
     setHiddenColumns(defaultTableConfig.hiddenColumns);
     setPageSize(defaultTableConfig.selectedPageSize);
     setShowMatchDetails(defaultTableConfig.showMatchDetails ?? false);
+    setAdvancedNestedSearch(defaultTableConfig.advancedNestedSearch ?? true);
```
DutchBen: Ask @wouterhhuisman but I think the default should be false?

packages/orchestrator-ui-components/src/components/WfoTable/WfoStructuredSearchTable/WfoFieldSelector.tsx
```diff
@@ -33,6 +33,7 @@ const focusValueEditorAfterRender = (searchInput: HTMLInputElement | null) => {
 export const WfoFieldSelector = ({ handleOnChange, disabled, rule, context }: FieldSelectorProps) => {
   const { field } = rule;
   const prefilledFieldOptions: FieldToOperatorMap = context.prefilledFieldOptions;
+  const advancedNestedSearch: boolean = context.advancedNestedSearch ?? true;
```
DutchBen: The name advancedNestedSearch doesn't indicate its a boolean. Ask Claude to come up with a name that does?

packages/orchestrator-ui-components/src/components/WfoTable/WfoStructuredSearchTable/WfoFieldSelector.tsx
```diff
@@ -43,13 +44,13 @@ export const WfoFieldSelector = ({ handleOnChange, disabled, rule, context }: Fi
     label: path,
   });
 
+  const isSelectablePath = (path: string) => advancedNestedSearch || !path.includes('.');
```
DutchBen: There should be a case for the column names - the onese that go above the ------. They should display in the list with their last .part only so it's still the columns of the table on the top of the choices.

DutchBen (approved):

PR #2699, opened 2026-08-20
222 query string

DutchBen:
Fixes the situation where you change something in either the filterString or the queryString and don't submit it right away (by 'enter') but submit the search by commiting from the other input (so either the query input field or the filterBuilder). In that case what get's submitted to search ignored those first changes leading to mistmatches to results and the url.

### DIFF
--- a/packages/orchestrator-ui-components/src/pages/WfoSearchPocPage.tsx
+++ b/packages/orchestrator-ui-components/src/pages/WfoSearchPocPage.tsx
@@ -344,23 +344,30 @@ export const WfoSearchPocPage = () => {
       sortableAndFilterableFieldNames,
     );
 
+  // Formats the given rule group to CEL and commits it. Returns false without committing when
+  // the CEL would not survive the URL round trip: formatQuery escapes double quotes in values
+  // but parseCEL has no escape support, so such a filter would silently be dropped after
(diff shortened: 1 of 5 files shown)


packages/orchestrator-ui-components/src/pages/WfoSearchPocPage.tsx
```diff
@@ -335,41 +339,64 @@ export const WfoSearchPocPage = () => {
...
+
+  const handleApplyFilter = (searchParams?: SearchParams) => {
+    const ruleGroupParam = searchParams?.ruleGroup;
+    // Use an explicitly passed rule group when provided (e.g. a column-header search), a cleared filter
+    // when `false`, and otherwise the current query builder state (the "Apply filter" button).
+    const effectiveRuleGroup = ruleGroupParam === false ? undefined : (ruleGroupParam ?? queryBuilderRuleGroup);
+    if (!commitFilterDraft(effectiveRuleGroup)) {
+      return;
+    }
+    // Also commit the draft search text, so applying a filter picks up text typed in the
```
tjeerddie, line 376: don't believe this comment adds much clatification then what the line already does

tjeerddie (approved):

wouterhhuisman (approved):

